Issue #10, opened 2015-04-01
How to use with props?


dubert:
If you wanted to include styles passed through props, what would be the best way to do it?

`useSheet()` is called while props are still undefined, so even passing through a method doesn't work without getting super hacky.

dubert:
I think I'm going to try to create my own [higher order component](https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750) that handles getting props and attaching them with JSS.

gaearon:
Please do! Also a PR would be nice. ;-)

gaearon:
I'd like to expose both the mixin and the HOC.

dubert:
I'm on it!

kof:
If you are not passing it too often, you can create an empty style like usually and then add rules to this style by using addRules method https://github.com/jsstyles/jss/blob/master/lib/StyleSheet.js#L160

kof:
But its better you explain your use case, the "why" behind what you are doing instead of what you trying to achieve .... because maybe you don't need it at all.

gaearon:
>But its better you explain your use case, the "why" behind what you are doing instead of what you trying to achieve .... because maybe you don't need it at all.

Agreed. Still we need HOC instead of mixin now that 0.13 is out.

dubert:
I have a component, and I want to be able to pass it css properties, like `color: #123456` and have them overwrite the components base style. I'm trying to do this because the css properties are variables, and so making various classes isn't an option.

kof:
why don't you use reacts inline styles for those overrides?

dubert:
I'm using pseudo elements: `&:hover`, `&:focus`, `&::-webkit-scrollbar`. I know I can do hover, and focus with onMouseEnter, and onFocus, but I hate making states for them when they can be done so easily in CSS. And they don't solve everything.

timbur:
What's the status on the HOC?  I'd like to use `react-jss` as a decorator, though my approach when converting mixins to decorators involves monkey-patching the component instead of wrapping it.  I am curious as to how one might turn this into a HOC.

Also lol... @gaearon you are everywhere man.  Today I've been sorting through all the different CSS-in-JS approaches I have bookmarked and concluded `jss` is the best one to go with.  So I check out the JSS directory and lo and behold, there's @gaearon in the upper right corner haha.  And @kof awesome work with JSS btw!  It's definitely the most logical, sensible and reusable approach.

[...]

PR #11, opened 2015-05-06
Added decorator (uses ES6+ composition)

timbur:
For #10


kof:
looks good to me @gaearon

gaearon:
Any particular reason for not making it a HOC? I'd rather avoid monkey patching whenever possible.

timbur:
I have no experience with creating HOC, but I guess today is as good a time as any.  I need to finish some other things at the moment but afterwards I can try to figure it out.  Any tips on how to make a HOC for this particular functionality would be great!

timbur:
I have no idea if I implemented it 100% correctly, but it seems to work as a HOC now.  Check the minor differences in the README.

I created [`react-mixin-decorator`](https://github.com/timbur/react-mixin-decorator) to make it really easy to either 1) convert existing mixins to ES7 decorators or 2) create ES7 decorators with fallback mixin support.  This PR uses this, of course, as a proof of concept.

gaearon:
To be honest I don't see the point of having a HOC/decorator that also acts as a mixin. It seems like artificial complexity because, if you have a HOC, you can wrap both new-style and old-style classes into it.

gaearon:
I'll try to come up with something similar to what you wrote, but without the mixin support.

gaearon:
I'm closing this in favor of #12, but thanks a lot for the effort, it helped me see a few things I'd like to change in 1.0.

PR #12, opened 2015-05-10
Export a higher-order component instead of a mixin

gaearon:
This is my take on a HOC. I tried to go with #11 but there were a number of things I wanted to differently so I went with a separate PR. (Still, thanks to @timbur for the effort, it helped me to see what I'd like to be changed :+1: )

This is what's different

* I'd like to avoid exporting both mixin and HOC. We're bumping the major version so it's not a big deal to stop supporting the mixin. People using `React.createClass` can still use the HOC—it will work for everyone so it's a better solution.

* I saw no reason for keeping `sheet` in the `state` so I keep it in an instance field instead, just like before.

* I'm removing `classSet` helper. I don't think it's appropriate now that React is going to deprecate their own `classSet` helper. There are microlibraries like [classnames](https://github.com/JedWatson/classnames) that do a better job on this, and we should just tell people to use them.

I'm going to update the PR with docs and release this later today.
Thanks again, @timbur and @kof.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,57 +1,170 @@
 ## React JSS
 
-This mixin makes [JSS](https://github.com/jsstyles/jss) easy to use from React components.
-Classes are always namespaced so use `this.sheet.classes` to access their names.
-Stylesheet is only attached until last instance is unmounted.
+Use this [higher-order component](https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750) to inject [JSS](https://github.com/jsstyles/jss) stylesheets in your React components. The stylesheet is attached when there is at least one mounted component that uses it, and automatically detached when all components using it are unmounted. React JSS is compatible with live reloading using [React Hot Loader](https://github.com/gaearon/react-hot-loader).
 
-This mixin is compatible with live reloading via [React Hot Loader](https://github.com/gaearon/react-hot-loader).
 
-### Example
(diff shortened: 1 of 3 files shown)


gaearon:
@kof Please review before I merge or publish.

gaearon:
Btw what do you think about renaming `this.props.sheet` to `this.props.jss`? Sounds more specific to me.

timbur:
Yeah that does make more sense.  As for my reasoning for leaving mixin support in #11, it was so people could be really lazy about updating their existing code.  The `react-mixin-decorator` I created also makes it quicker and easier (in my opinion) to create HOCs as decorators since everything in that area becomes standardized and predictable, with the HOC's defaultProps/state automatically passed as props and its methods bound.

gaearon:
We have no new features in 1.0 so people don't need to update if they don't want to. I don't think a smoother upgrade path is worth the underlying complexity in this case.

I see your reasoning but is makes it significantly harder to understand what's going on, and don't see much value in making this tradeoff. People can just stay on 0.x for the old code they don't want to change. If we don't do this now, it will be even harder to drop the mixin in the future.

timbur:
Yeah, no problem.  I included the `react-mixin-decorator` with PR #11 mainly as an experiment, to see if it would actually work.  IMO, it is kind of nice to be able to quickly create HOCs that can double as mixins at no extra cost or effort for different people who may have different project requirements or if they simply prefer the mixin pattern.  Just something I was trying out!  :)

gaearon:
I don't like this approach because people are confused enough about decorators already. Seeing something that is both a mixin and a decorator, I'm afraid, is only going to add to the confusion, and discourage people from actually understanding the underlying pattern (which is simple and powerful).

gaearon:
I added a friendlier ES5/ES6 overload that takes React component class as the first parameter. I think it's a great tradeoff, as it makes the lib clean and easy to use from both ES5, ES6 and ES7.

PR #17, opened 2015-07-27
Added a ref value to the DecoratedComponent

dwighthouse:
React-Jss has moved from a mixin system to a higher-order component system. Within it, the real component is wrapped in another component that contains jss-specific functionality and then exposes the new functionality in the form of an injected prop called 'sheet'. This normally works because the react-jss wrapping component (hereunto called 'StyleSheetWrapper') propagates the props forward into the real component in its render method.

However, there is another form of inter-component communication that is relatively rare, but important called [Exposed Component Functions](https://facebook.github.io/react/tips/expose-component-functions.html). Using a ref to a child component, a parent component may directly call a method on the child component. This wouldn't be used to transfer data, but it would be used to trigger actions on the child from the parent not reasonably expressed with data, such as "You've just become 'active,' focus the cursor to your appropriate input field."

By merging this pull request, a parent component can use the following code to call a function on a child component that has been wrapped by react-jss (and only react-jss).
`this.refs.WRAPPED_COMPONENT.refs.wrapped.CHILD_METHOD()`
as opposed to the following when using mixins
`this.refs.CHILD_COMPONENT.CHILD_METHOD()`

This solution leaves a lot to be desired, though these issues are not specific to react-jss, but to higher-order component usage by React in general.

First, StyleSheetWrapper might be wrapping another wrapped component, or it might be wrapped itself. There is no way for StyleSheetWrapper to know where the original component is, nor any other wrapping higher-order component in the chain. Furthermore, other wrapping components might not wrap just one component, but may instead create meaningful refs that are not the original component.

Second, while replacing mixins with higher-order components removes the potential for multiple inheritance collisions, now the parent must have knowledge that the child is wrapped, in what order it is wrapped, and what refs those wrappers create.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -61,7 +61,7 @@ function decorate(DecoratedComponent, rules, options = {}) {
 
     render() {
       return (
-        <DecoratedComponent {...this.props} sheet={this.sheet} />
+        <DecoratedComponent ref="wrapped" {...this.props} sheet={this.sheet} />
       );
     }
   };


kof:
The problem with this approach is that you never know when you use some component how often it has been wrapped, accessing a child this way may be a trip.

Also did you see this discussion? https://github.com/jsstyles/react-jss/issues/15

dwighthouse:
I had not seen that! You guys already figured out the limitations and drawbacks.

Reading it over, it looks like the recommended solution is to pass a function via props that is called by child code on startup to give the parent back a reference to its internal method, since props are the only things propagated throughout all wrapping components. It appears to work. Could we get documentation on this?

kof:
well, this is not react-jss specific ...

dwighthouse:
Here's a complete working example of the function-pass-by-props relative to wrapping with components such as react-jss using ES5 so there's no confusion with newer concepts. Feel free to use or modify for documentation.

    // Parent.js
    'use strict';

    var React = require('react');
    var Child = require('./Child');

    module.exports = React.createClass({
        click: function() {
            this.focusFirstInput();
        },
        setFirstFocusInput: function(focusFunc) {
            this.focusFirstInput = focusFunc;
        },
        render: function() {
            return (
                <div>
                    <h3 onClick={this.click}>Name</h3>
                    <Child setFocusFunc={this.setFirstFocusInput} />
                </div>
            );
        }
    });

    // Child.js
    'use strict';

    var React = require('react');
    var reactJss = require('react-jss');

    var Child = React.createClass({
        focusInput: function() {
            React.findDOMNode(this.refs.firstInput).focus();
        },
        componentDidMount: function() {
            this.props.setFocusFunc(this.focusInput);
        },
        componentWillUnmount: function() {
            this.props.setFocusFunc(() => {});
        },
        render: function() {
            var classes = this.props.sheet.classes;

            return (
                <div>
                    <label className={classes.label}>First Name</label>
                    <input ref="firstInput" className={classes.input} type="text" />

                    <label className={classes.label}>Second Name</label>
                    <input ref="secondInput" className={classes.input} type="text" />
                </div>
            );
        }
    });

    var styles = {
        label: {
            display: 'block'
        },
        input: {
            padding: '6px 10px',
            border: '2px solid #555'
        }
    };

    module.exports = reactJss(Child, styles);

kof:
> react-jss specifically breaks a part of React's documented feature set. 

Sorry maybe I missed something, can you explain please?

dwighthouse:
By using react-jss (or any wrapping component), the user will immediately loose the ability to use a ref to access [Exposed Component Functions](https://facebook.github.io/react/tips/expose-component-functions.html), a documented feature of React's API.  For users who need that feature, or its equivalent, there is no mention on react-jss's readme that the feature is disabled, why the feature is disabled, or what workarounds are available.

kof:
Well, as you also said its not react-jss specific ... I don't think we need to introduce workarounds for this as part of react-jss ... There should be some better place for this. Like blog posts with more detailed explanation. But I am happy to link to such a blog post.

kof:
@gaearon do you know a good article about decorators and accessing decorated components in react?

kof:
@dwighthouse I have just got to understand the real issue here. The point of the issue which is react-jss specific is that react-jss returns a new component. Basic decorator would return the same component back and not introduce this issue.

kof:
I think its possible that there is no article about this specific situation .... and somebody will need to write one, I will try to find time in the comming week to do it.

dwighthouse:
Oh, I thought you had understood, but still considered it unnecessary. You are right. There is no article I know of that addresses this scenario, just a couple of stack overflow comments and this ticket.
Feel free to use the above example code to demonstrate when you write about it.

PR #19, opened 2015-09-14
Extend wrapper class with React.Component

alexkuz:
stateless components in React v0.14.0 require components to extend React.Component.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -36,7 +36,7 @@ function decorate(DecoratedComponent, rules, options = {}) {
     DecoratedComponent.name ||
     'Component';
 
-  return class StyleSheetWrapper {
+  return class StyleSheetWrapper extends React.Component {
     static wrapped = DecoratedComponent;
     static displayName = `JSS(${displayName})`;
 


kof:
can you point me to the description or code to this change in react btw do you know why the need for extend was added?

gaearon:
From what I remember, this is needed because React 0.14+ supports stateless components that are pure functions. React needs to be able to tell a function from a class, because it will `new` a class, but arrow functions are not `new`able. Right now they use prototype check, but in 0.15 extending `Component` will be mandatory for class-like behavior.

I'm afraid I don't understand the specifics; @spicyj or @sebmarkbage might help clarify.

gaearon:
OK, found it:

https://github.com/facebook/react/issues/4599https://github.com/facebook/react/pull/4663

kof:
Ah so thats the way how they differentiate stateless function components from class components, I see.

kof:
merged, thanks

alexkuz:
👍

PR #23, opened 2015-11-27
Update .babelrc instructions to correct stage and babel 6 format

cheapsteak:
Decorators are now stage 1
Change to use babel 6 format


kof:
thanks

PR #33, opened 2016-03-23
Modified README

JeremyMiller:
README examples wouldn't work for me. It seems like `vendorPrefixer` is a function that must be invoked.


kof:
thanks

PR #36, opened 2016-04-28
Fixed mention of babel stage for decorators by the babel-plugin-transform-decorators-legacy plugin

adriantoine:
Hi,

Babel doesn't support decorators anymore: 
- http://babeljs.io/docs/plugins/transform-decorators/
- http://babeljs.io/docs/plugins/preset-stage-1/

So you need to install https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy to use decorators now.

I'm just fixing the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ class Button extends Component {
 export default useSheet(Button, styles)
 ```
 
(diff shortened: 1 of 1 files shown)


kof:
thanks

Issue #38, opened 2016-06-14
Hoist Statics from DecoratedComponent 


umidbekkarimov:
Use [hoist-non-react-statics](https://github.com/mridgway/hoist-non-react-statics)

kof:
released

PR #39, opened 2016-06-14
Hoist Statics from DecoratedComponent

umidbekkarimov:
Fix for #38 

I had problems while running `build` locally so I updated and add all dev dependencies to `package.json`. Also I slightly fixed `build` script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "description": "JSS mixin for React components",
   "main": "lib/index.js",
   "scripts": {
-    "build": "./node_modules/.bin/cross-env NODE_ENV=production BABEL_ENV=commonjs ./node_modules/.bin/babel src --out-dir lib",
+    "build": "cross-env NODE_ENV=production BABEL_ENV=commonjs babel src --out-dir lib",
     "prepublish": "npm run lint && npm run build",
-    "lint": "./node_modules/.bin/eslint ./src"
+    "lint": "eslint ./src"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -48,5 +48,8 @@
     "jss": "^3.11.1",
     "react": "^15.1.0",
     "webpack": "^1.13.1"
+  },
+  "dependencies": {
+    "hoist-non-react-statics": "1.1.0"
   }
 }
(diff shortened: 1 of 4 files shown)


kof:
I am not sure copying static props to the HOC is a good thing. Need more input for pros and cons.

umidbekkarimov:
@kof 

> I am not sure copying static props to the HOC is a good thing.

Some powerful libs that provide HOC and use [hoist-non-react-statics](https://github.com/mridgway/hoist-non-react-statics) in it:

* react-redux - [@connect](https://github.com/reactjs/react-redux/blob/master/src/components/connect.js#L365) 
* react-router - [@withRouter](https://github.com/reactjs/react-router/blob/master/modules/withRouter.js#L20)
* redux-form - [@reduxForm](https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L39)

>  Need more input for pros and cons.

I have `SelectField` component with it's static method `SelectField.getOptionKey` which is used like simple helper to obtain value by key.

I used it in this way:

```javascript

import SelectField from './SelectField';

function PackageSelectField({
    packagePrices,

    fields: {
        packageItem
    }
}) {

    return (
        <SelectField
            {...packageItem}
            options={packagePrices}
            hintText="Choose Package"

            getOptionText={SelectField.getOptionKey('text')}
            getOptionValue={SelectField.getOptionKey('value')}
        />
    );
}

```

package.json
```diff
@@ -4,9 +4,9 @@
   "description": "JSS mixin for React components",
   "main": "lib/index.js",
   "scripts": {
-    "build": "babel src --out-dir lib",
+    "build": "./node_modules/.bin/cross-env NODE_ENV=production BABEL_ENV=commonjs ./node_modules/.bin/babel src --out-dir lib",
```
kof: full path should not be needed, why? npm adds it automatically ...

umidbekkarimov: Really didn't know that, I had problem with it in `travis-ci` once so i add full path every time.

kof:
looks good

kof:
don't forget to lint

kof:
huh, after updating dependencies linter stopped having that weird error.

umidbekkarimov:
@kof yes, i updated `eslint` version because of strange errors

kof:
published

PR #46, opened 2016-08-09
Change JSS peerDependency from 5.2.0 to ^5.2.0

dlwalsh:
I can see no reason why the jss peer dep should be fixed at 5.2.0.


kof:
thanks, thats a typo.

PR #53, opened 2016-11-07
Theme support using context variables

jcperez-ch:
- Basic skin implementation for components using composition.
- Added unit tests U
- Updated README file. 

Please review and let me know your comments:

```javascript
import React, {Component} from 'react'

@injectSheet(parentProps => {
  const {
    buttonBackgroundColor = 'yellow',
    labelWeight = 'normal'
  } = parentProps;

  return {
    button: {
      backgroundColor: buttonBackgroundColor
    },
    label: {
      fontWeight: labelWeight
    }
  }
})
export default class Button extends Component {
  render() {
    const {sheet: {classes}, children} = this.props
    return (
      <button className={classes.button}>
        <span className={classes.label}>
          {children}
        </span>
      </button>
    )
  }
}
```


kof:
cc @nathanmarks

kof:
Lots of things to be considered, I want a solid theming solution.

kof:
Have you seen https://github.com/nathanmarks/jss-theme-reactor ?

jcperez-ch:
I understand about the need of a whole theming/skinning solution.  This is more like a meanwhile placeholder that allows us to somehow skinning reactively.  `jss-theme-reactor` seems more like a manager, which would be a good idea when considering skinning.  Thanks for the feedback ;)

kof:
Question: would your use case be solved [this way](https://github.com/cssinjs/react-jss#using-classnames-helper)?

jcperez-ch:
@kof Indeed.  I am actually using it this way, works really nice.  Did you try?

kof:
> Did you try?

Did I try classes composition? Sure, thats what I do all the time.

jcperez-ch:
Oh sorry, I completely interpreted wrong the question. No, using `classNames` didn't solve my problem.  I wanted my component to react to the props of the component, where there would be logic that defines the skinned style object.

kof:
What would happen if your component renders multiple times with different props?

kof:
A sheet produced by react-jss is shared between all elements of a given component. This means that if a component can rerender the sheet, it neeeds to render a different sheet, it can't rerender the one shared between all elements.

By element I mean the result of React.createElement().

jcperez-ch:
I kept the `refs` counter that allows a jss instance to be attached per element, so suppose we have a List component with decorated Item components

If for some reason List renders this:

```js
<List>
  <Item skin="default" />
  <Item skin="other" />
</List>
```

the classes defined by `default` will attach the style, so the second item won't get new styles, so this approach *won't work* since the decorated Item would react `onComponentUpdate`, so, if the list re-renders to this:

```js
<List>
  <Item skin="first" />
  <Item skin="second" />
</List>
```

Then the first item would detect a change to the `first` skin param, then the `second` would detect a new change, so at the end both items would have the `second` skin styles.  Obviously that is not ideal neither.

So:
- There would be one skin for the elements of a class at all time no matter if all instances have different props that define the skin.
- JSS won't attach a different style tag per skin and per element. I use the `jss.removeStyleSheet(sheet)` if a change is detected.

kof:
We can't do that. Its totally not what one would expect from this api.

kof:
If I render this, as a user, I expect both components styled properly.

```javascript
      const WrappedComponent = injectSheet(props => ({
        button: {color: props.buttonColor || 'red'}
      }))()
      render(<div><WrappedComponent /><WrappedComponent buttonColor="green" /></div>, node)
```

jcperez-ch:
Yeah, that is not gonna happen with the approach of this PR since the problem I wanted to solve was the skinning and the reaction of all WrappedComponents via a shared skin prop used globally, if you want a button blue, you see it blue in every implementation (so in every React.createElement).

A better approach is to have a `SkinProvider` using a childContextType param that defines the skin (similar to what react-redux does for the store)... also we can leverage the use of @nathanmarks's [jss-theme-reactor](https://github.com/nathanmarks/jss-theme-reactor)  so, the Wrapper Component inside the decorator will react on the context value and we can injectStyles like: 

```js
injectSheet(theme => ({
  button: {color: theme.buttonColor || 'red'}
}))
```

So you're right about it. in the meanwhile my approach solved the problem I had, but it is prune to misinterpretation of the actual solution.  I'll see the jss-theme-reactor soon and work in it too.

jcperez-ch:
@nathanmarks @kof I updated the code so instead of props it uses the `theme` context.  We can encourage to use a Provider that works well with [jss-theme-reactor](https://github.com/nathanmarks/jss-theme-reactor).

Read the entry I left for the [README](https://github.com/jcperez-ch/react-jss/blob/composed-styles/README.md).

kof:
Next is this one.

jcperez-ch:
The support for a provider is very limited with this approach given the fact that I don't cover:

- The used provider.
- The fact that context is never propagated into pure components.

At the end I resolved my problematic with a custom implementation leveraging redux and extending `jss` outside `jss-react`.  I believe this PR won't be relevant anymore if there is a jss theme provider plugin.  @kof what are your thoughts?

jcperez-ch:
Closing, since skin resolution would be on the side of `jss` not on the side of `reacct-jss`

PR #56, opened 2016-11-12
create JssSheetsRegistry component

jedwards1211:
Note: this depends on https://github.com/cssinjs/jss/pull/355

It puts a `SheetsRegistry` from its props onto its child context as `jssSheetsRegistry`.

Then any `injectSheets` wrapper components check if `jssSheetsRegistry` is on their `context`, and if so, they add/remove their sheets from it.

I added a test and example usage for server-side rendering to the README.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,41 @@ const Component = ({sheet: {classes}, children, isActive}) => (
 )
 ```
 
+### Using `JssSheetsRegistry` for server-side rendering
+
+```es6
+import {renderToString} from 'react-dom/server'
+import {SheetsRegistry} from 'jss'
+import {JssSheetsRegistry} from 'react-jss'
+import MyApp from './MyApp'
+
+export default function render(req, res) {
+  const sheets = new SheetsRegistry()
+
+  const body = renderToString(
+    <JssSheetsRegistry registry={sheets}>
+      <MyApp />
+    </JssSheetsRegistry>
+  )
+
+  // any instances of `injectStyle` within `<MyApp />` will have gotten `sheets`
+  // from `context` and added their style sheets to it by now.
+
+  return res.send(renderToString(
+    <html>
+      <head>
+        <style type="text/css">
+          {sheets.toString()}
+        </style>
+      </head>
+      <body>
+        {body}
+      </body>
+    </html>
+  ))
+}
+```
+
 ### Installation.
 
 ```
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-jss",
-  "version": "4.1.3",
+  "version": "4.2.0",
   "description": "Integration of JSS for react.",
   "main": "lib/index.js",
   "engines": {
--- a/src/JssSheetsRegistry.js
+++ b/src/JssSheetsRegistry.js
@@ -1,4 +1,4 @@
-import {Component, PropTypes} from 'react'
+import React, {Component, PropTypes, Children} from 'react'
 import {SheetsRegistry} from 'jss'
 
 export default class JssSheetsRegistry extends Component {
@@ -15,7 +15,12 @@ export default class JssSheetsRegistry extends Component {
     }
   }
   render() {
-    return this.props.children || null
+    const {children} = this.props
+    return (
+      Children.count(children) > 1
+        ? <div>{children}</div>
+        : children
+    )
   }
 }
 
--- a/src/JssSheetsRegistry.js
+++ b/src/JssSheetsRegistry.js
@@ -1,0 +1,21 @@
+import {Component, PropTypes} from 'react'
+import {SheetsRegistry} from 'jss'
+
+export default class JssSheetsRegistry extends Component {
+  static propTypes = {
+    registry: PropTypes.instanceOf(SheetsRegistry).isRequired,
+    children: PropTypes.node.isRequired
+  }
+  static childContextTypes = {
+    jssSheetsRegistry: PropTypes.instanceOf(SheetsRegistry).isRequired
+  }
+  getChildContext() {
+    return {
+      jssSheetsRegistry: this.props.registry
+    }
+  }
+  render() {
+    return this.props.children || null
+  }
+}
+
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,10 @@
-import React, {Component} from 'react'
-import {create as createJss} from 'jss'
+import React, {Component, PropTypes} from 'react'
+import {create as createJss, SheetsRegistry} from 'jss'
 import preset from 'jss-preset-default'
 import hoistNonReactStatics from 'hoist-non-react-statics'
+import JssSheetsRegistry from './JssSheetsRegistry'
+
+export {JssSheetsRegistry}
 
 /**
  * Wrap a Component into a JSS Container Component.
@@ -46,9 +49,14 @@ function wrap(jss, WrappedComponent, styles, options = {}) {
   return class Jss extends Component {
     static wrapped = WrappedComponent
     static displayName = `Jss(${displayName})`
+    static contextTypes = {
+      jssSheetsRegistry: PropTypes.instanceOf(SheetsRegistry)
+    }
 
     componentWillMount() {
       this.sheet = ref()
+      const {jssSheetsRegistry} = this.context
+      if (jssSheetsRegistry) jssSheetsRegistry.add(this.sheet)
     }
 
     componentWillUpdate() {
@@ -64,6 +72,8 @@ function wrap(jss, WrappedComponent, styles, options = {}) {
     componentWillUnmount() {
       if (this.sheet && !sheet && !refs) {
         this.sheet.detach()
+        const {jssSheetsRegistry} = this.context
+        if (jssSheetsRegistry) jssSheetsRegistry.remove(this.sheet)
       }
       else deref()
     }
--- a/src/index.test.js
+++ b/src/index.test.js
@@ -3,7 +3,7 @@ import jss, {create as createJss, SheetsRegistry} from 'jss'
 import React from 'react'
 import {render, unmountComponentAtNode} from 'react-dom'
 import deepForceUpdate from 'react-deep-force-update'
-import injectSheet, {create as createInjectSheet, jss as reactJss} from './'
+import injectSheet, {create as createInjectSheet, jss as reactJss, JssSheetsRegistry} from './'
 
 const node = document.createElement('div')
 
@@ -184,22 +184,25 @@ describe('react-jss', () => {
 
   describe('with JssSheetsRegistry', () => {
     it('should add style sheets to the registry from context', () => {
-      class AppContainer extends React.Component {
-        render() {
-          const sheets = new SheetsRegistry()
+      const sheets = new SheetsRegistry()
 
-          return (
-            <JssSheetsRegistry registry={sheets}>
-              <WrappedComponentA
-                {...this.props}
-                key={Math.random()} // Require children to unmount on every render
-              />
-            </JssSheetsRegistry>
-          )
+      const Component = () => null
+      const WrappedComponentA = injectSheet({
+        button: {color: 'red'}
+      })(Component)
+      const WrappedComponentB = injectSheet({
+        button: {color: 'blue'}
+      })(Component)
 
-          expect(sheets.registry.length).to.equal(2)
-        }
-      }
+      render(
+        <JssSheetsRegistry registry={sheets}>
+          <WrappedComponentA />
+          <WrappedComponentB />
+        </JssSheetsRegistry>,
+        node
+      )
+
+      expect(sheets.registry.length).to.equal(2)
     })
   })
 })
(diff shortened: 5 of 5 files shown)


kof:
I think we made a solid theming concept https://github.com/cssinjs/jss/issues/356

README.md
```diff
@@ -117,6 +117,41 @@ const Component = ({sheet: {classes}, children, isActive}) => (
...
+import {renderToString} from 'react-dom/server'
+import {SheetsRegistry} from 'jss'
+import {JssSheetsRegistry} from 'react-jss'
+import MyApp from './MyApp'
+
+export default function render(req, res) {
+  const sheets = new SheetsRegistry()
+
+  const body = renderToString(
+    <JssSheetsRegistry registry={sheets}>
```
kof: Why do we need to pass the registry?

kof: Can't `JssSheetsRegistry ` create one for us?

kof: Oh, you need it because you want to make sheets.toString().

kof: The problem I see here is that jss library is actually an internal thing for react-jss, it should expose everything one needs, otherwise it is confusing for lots of people. I have been there already.

kof: Another problem is that you are actually introducing a second sheets registry, the first one is still in the jss instance and will continue aggregating all sheets for every request!

jedwards1211: @kof but if `injectSheet` falls back to `jss` passed as an argument will it not add the sheets to the registry from `JssProvider`?

jedwards1211: @kof sorry nevermind about that, but there's a bigger problem: if all `injectSheets` are using the one `jss` from `JssProvider`, they all depend on the same plugin config.  I designed this PR so that all components could still use their own jss plugin config if necessary

jedwards1211: The registry is not used for client-side rendering is it?  I think having both the registry and plugin config in a `jss` instance is problematic for server-side rendering

jedwards1211:
@kof the purpose of this PR isn't theming, I'm just trying to make it possible to do SSR without global variables.

README.md
jedwards1211: Speaking of which, even with the current design it's not possible to get the styles for components using local jss in server-side rendering, is it?

jedwards1211: and that's true that the first jss instance would continue aggregating sheets on every request.  That's why I wish there were a separation between the plugin config and the registry.

kof: I was thinking about that. I actually start like your points. Right now I even think to remove the default SheetsRegistry from jss. As far as I know this can be needed only for SSR anyways. I think it should be an option and should not be set by default. Because if nobody cares, why should it aggregate the sheets anyways?

cc @nathanmarks

kof: This way also it is more clear to the user that he needs to create a new registry on every request and we are more on a secure side in terms of memory leaking.

jedwards1211: Yeah, my thoughts exactly :)

jedwards1211: @kof are you waiting on anything from me for this?

kof: Nope, I am busy with 2 other major things, will ping you on this one asap.

kof:
Do you see any problems if `const jss = create({sheets: new SheetsRegistry()})` and `jss.createStyleSheet()` adds automatically the sheet to the registry if one has been passed. Default is no registry at all?

its  a part of a bigger PR for the next major version https://github.com/cssinjs/jss/pull/348

jedwards1211:
I don't see any problems, that sounds like a good way to support the old way of doing SSR

kof:
Just realized its actually a problem, because this way you can't aggregate sheets in a registry from all jss instances.

kof:
Another question: what if you use an external component which doesn't use react-jss, how would you aggregate sheets for SSR?

kof:
Maybe it is completely out of scope for JSS, it should be probably an API decision of a component.

kof:
An external component might also not be using JSS at all, you still are going need to aggregate styles for SSR.

jedwards1211:
@kof Oh you were suggesting `const jss = create({sheets: new SheetsRegistry()})` instead of this PR?  I thought you meant it would be an additional option.

Right, passing the sheets registry into `create` wouldn't be sufficient to aggregate all the sheets from 3rd party components.  This PR is still the best way I can think of to do that.

If we merge this PR it wouldn't prevent people from using other solutions anyway.  If external components didn't want to depend on `react-jss`, they could declare the registry in their `contextTypes` and use it.  They could also allow you to pass in the registry as a prop.

jedwards1211:
@kof if an external component is using something besides JSS, whatever library it's using will have to provide its own way of getting its sheets into SSR.  But I don't see why that would interfere with JSS aggregating its sheets into a registry on context.

jedwards1211:
@kof think about how redux and react-router work -- both pass down global state via context.  I think that pattern has proven itself flexible and robust.

kof:
my prob right now in jss core is one feature we had before, which relies on SheetsRegistry (index based sheets ordering). So if one wants to use the index option, sheets need to be passed. Then its a question wether jss should add a  sheet to the registry on jss.createStyleSheet or not.

kof:
Not doing it automatically means for someone who uses the index option and doesn't use react-jss or other abstractions on top, that he will need to add the sheet after every createStyleSheet. Which is a lot of additional typings for the user.

kof:
Also if user is using react-jss + index option, you still need to pass registry to Jss, because otherwise index option won't work.

jedwards1211:
I looked through the code and I don't understand why the index option wouldn't work.  The index is stored in the sheet and the registry handles the index, so it doesn't seem to depend on what jss instance is used.

jedwards1211:
@kof I understand it's additional typing over the current design but remember:
* you could use `react-jss` with this PR to avoid that typing
* The current design *can't aggregate stylesheets from components that create their own jss*

jedwards1211:
@kof the other problem with passing a registry in when creating JSS is that typically you create JSS in module scope instead of once per SSR.

jedwards1211:
Oh -- on the client side, is a registry necessary for `index` to work?

kof:
> Oh -- on the client side, is a registry necessary for index to work?

Yes, basically SheetsRegistry is used for 2 different things. One is needed for SSR only, the other one for both.

kof:
Maybe we just need to have  a global singleton of SheetsRegistry for all jss instances.

kof:
```javascript
import {sheets} from 'jss'
// Do this on every request.
sheets.reset()
```

jedwards1211:
@kof I thought you were also hoping to avoid having a global singleton?

jedwards1211:
I really don't understand why it's such a problem for the user to create a sheets registry and put it on context to use jss.  I don't think anyone avoided using redux because they were afraid of the extra typing that `<Provider>` and `connect` required.

kof:
> I really don't understand why it's such a problem for the user to create a sheets registry and put it on context to use jss. I don't think anyone avoided using redux because they were afraid of the extra typing that <Provider> and connect required.

Its not that. Its about jss being used without any wrapper like react-.jss or even not with react at all.

jedwards1211:
Hmmm, that's a good point, I keep forgetting that people might use it without react at all.  Lemme think on this for awhile then

jedwards1211:
The question is if there can be a clean way to pass a fresh SheetsRegistry for each request to non-react 3rd party libs that use jss

jedwards1211:
@kof the biggest problem with having a global registry is that if SSR is done asynchronously (for instance with react-dom-stream) then styles for one request can wind up in the response for another request.

I know that react-dom-stream doesn't work very well with css in js right now, since you would have to render the body as a string so that stylesheets are available, and then stream the styles and body in the boilerplate html, but I think async rendering will only become more common.

kof:
Yeah I also think its a bit opinionated to assume the sync rendering always.

jedwards1211:
Do you know of any examples of people using jss in non-react SSR?  Would be helpful to look at

kof:
Nope. I can tweet out, maybe I will get some.

jedwards1211:
I still think it would be best to eventually encourage non-react projects/libraries to create a way of passing in a different `SheetsRegistry` for each request.  But it would be helpful to see how compatible that would be with people's current approaches

jedwards1211:
Also, there's no reason we can't give users both options: use a single global `SheetsRegistry` or pass one down via react context, and add sheets to whichever (or both) is available.

kof:
Well the decision I need to make right now is how to keep the index option with that approach.

kof:
What if we use a function which returns a SheetsRegistry instance? On new request we set `sheets = null`. This way the entire tree can keep any amount of different jss instances, they all need however take the `getSheets` function.

````
import {create} from 'jss'

let sheets
const jss = create({
  getSheets: () => {
    if (!sheets) sheets = new SheetsRegistry()
    return sheets
  }
})
````

kof:
Ok, same bad idea, updating sheets will rely on some global state as before.

kof:
Ok I don't see any other way than to resort back to my initial proposal, passing a new Jss instance down the context and letting jss handle the sheets registry.

jedwards1211:
I don't think that's necessary...jss could check if a DOM is present, and if so, assume it's client side and create a global SheetsRegistry.  Then when a StyleSheet is created, automatically add it to that global SheetsRegistry if it exists.  Or the SheetsRegistry could be stored in the `DomRenderer`.

Remember, passing down a single jss instance means all components depend on having the same jss config.

kof:
Different back and client logic is an inconsistency from the api perspective. 

>Remember, passing down a single jss instance means all components depend on having the same jss config, which could be a problem.

We can workaround this.

kof:
One way to solve different jss setups in one tree but the same sheets registry:

JssProvider creates a jss instance. If there is already a jss instance in the context, it uses its registry to create a new jss instance which will overwrite the parent one.

jedwards1211:
> Different back and client logic is an inconsistency from the api perspective.

Actually I figured out how to implement it with a static SheetsRegistry on DOMRenderer instead of a global variable, so that it wouldn't affect the public API: https://github.com/cssinjs/jss/pull/373

jedwards1211:
> JssProvider creates a jss instance. If there is already a jss instance in the context, it uses its registry to create a new jss instance which will overwrite the parent one. In other words JssProvider can inherit the registry automatically.

But then all 3rd party components would have to wrap themselves in a `JssProvider`...more awkward than just creating a local `jss` instance I think.

kof:
> Actually I figured out how to implement it with a static SheetsRegistry 

There are 2 problems with this:
1. This will leak memory on the server because nobody cleans that static registry up, right?
2. Index option won't work globally across all components. I think a user would expect it globally (not sure though.)

jedwards1211:
1. But `DOMRenderer` isn't supposed to be used on the server, right?
2. Why wouldn't indexes work globally?  There's only one sheets registry, all components' styles get put into it.

kof:
> But DOMRenderer isn't supposed to be used on the server, right?

true, forgot that

> Why wouldn't indexes work globally? There's only one sheets registry, all components' styles get put into it.

true, I oversaw that the static property is created just once.

kof:
You actually just discovered another problem: index option is implemented in dom renderer and won'T work server-side. This is actually a fat bug.

kof:
For universal apps its a deal breaker. The order should always be the same.

kof:
Actually it works, we just have no tests, because registry ensures correct index.

kof:
Hmm will sleeep over it

jedwards1211:
Cool, glad I was able to explain it better!

jedwards1211:
> Actually it works, we just have no tests, because registry ensures correct index.

But with the current design, if the user creates multiple jss instances, each has its own separate registry -- so sheets would not be sorted globally on the client or server; they would only be in order relative to other sheets in the same registry.

jedwards1211:
and actually instead of using a static properly, maybe I should just use a module-scope variable for the sheets in DOMRegistry, so that it's not publicly accessible

kof:
Ok done, check this out https://github.com/cssinjs/jss/pull/348/commits/ad0ab92f5ca2ff6352644d1aae2f69b56ad86847

kof:
Differences to your initial PR on jss repo are:
- sheets registry is create in index.js, it is public
- sheets are always added to the registry when DomRenderer is created
- .toString of SheetsRegistry stringifies only  attached sheets.

jedwards1211:
Sounds good!

kof:
JSS 6 is released, now we can revisit this.

kof:
We got a conflict here.

jedwards1211:
Okay, I fixed the conflicts

kof:
merged

kof:
Released with a few changes, see changelog.

jedwards1211:
Just integrated the new `SheetsRegistryProvider` into my project, it works really well!  Thanks for your patience on this!

jedwards1211:
Do you want me to PR an update to the SSR example code in `jss`?

kof:
Sure, any help is appreciated!

Issue #61, opened 2016-12-13
Source order


kof:
By design, React renders parents first, children second. This results in sheets rendered from parent first, child second. When you pass a class name from the parent to the child, childs style is later in the head, so it has a higher source order specificity and you can't override it from the parent without using '!important'.

Now we need to think of a way to reverse order for sheets rendering: child's sheet first, parents sheet second.

cc @nathanmarks

cvle:
You can solve this by introducing a `createStyles` function, that assigns an index to the styles using a global index counter.

~~~js
import React from 'react'
import injectSheet from 'react-jss'

let globalIndex = 0

function createStyles(styles) {
  return { 
    styles,
    index: globalIndex++
  }
}

const styles = createStyles({
  button: {
    backgroundColor: 'yellow'
  },
  label: {
    fontWeight: 'bold'
  }
})

const Button = ({sheet: {classes}, children}) => (
  <button className={classes.button}>
    <span className={classes.label}>
      {children}
    </span>
  </button>
)

// injectSheet uses index when creating the style sheet.
export default injectSheet(styles)(Button)
~~~

Naturally this is a breaking change to the current API.

kof:
lets go for it!

kof:
released and merged, thanks!!!!

PR #62, opened 2016-12-20
[WIP] Preserve source order

cvle:
Just had an idea to solve #61 without changing the API.
@kof if you agree with this, I'll finish the PR by adding a test.


kof:
We thought about reverse order rendering of styles with @nathanmarks and he found a case where this still breaks.

kof:
It is fixing a simpler case though.

kof:
Here is the case where its still a problem.

```
class MyAwesomeApp {
  render() {
    <div>
      <MyAwesomeIcon />
      <h1>WELCOME!</h1>
      <MyAwesomeToolbar />
    </div>
  }
}

class MyAwesomeToolbar {
  render() {
    const classes = this.context.styleManager.render(styleSheet);
    return (
      <div className={classes.root}>
        <MyAwesomeIcon className={classes.icon} />
      </div>
    );
  }
}
```

kof:
I don't see any real fix right now other than "never pass a class from a parent"

cvle:
I don't really understand your issue yet and what you exactly mean by "reverse order rendering of styles". The solution I propose is to keep a global index counter to set the index at the same point as when the source code was imported / parsed. This way we have the same behavior as when we were using CSS-Modules.

So in this case your example would be:

~~~javascript
@injectSheet(stylesA) // index is set to -10000 and used later for style sheet creation
class MyAwesomeIcon {
  render() {
    const classes = this.props.sheet.classes;
    return (
      <div className={classes.root + " " + this.props.className}></div>
    );
  }
}
~~~

~~~javascript
import {MyAwesomeIcon} from "./MyAwesomeIcon";

@injectSheet(stylesB) // index is set to -9999 and used later for style sheet creation
class MyAwesomeToolbar {
  render() {
    const classes = this.props.sheet.classes;
    return (
      <div className={classes.root}>
        <MyAwesomeIcon className={classes.icon} />
      </div>
    );
  }
}
~~~

~~~javascript
import {MyAwesomeIcon} from "./MyAwesomeIcon";
import {MyAwesomeToolbar} from "./MyAwesomeToolbar";

@injectSheet(stylesC) // index is set to -9998 and used later for style sheet creation
class MyAwesomeApp {
  render() {
    <div>
      <MyAwesomeIcon />
      <h1>WELCOME!</h1>
      <MyAwesomeToolbar />
    </div>
  }
}
~~~

Resulting in the following priority order: 

MyAwesomeApp -> MyAwesomeToolbar -> MyAwesomeIcon

cvle:
Actually this behavior is even better than with CSS Modules. Because in CSS Modules the order is determined upon import, but this proposal would determine order upon Component Definition. This is actually significant because this is the most optimal time to do so. With CSS Modules the style sheet order can get wrong when dependent styles are imported between the style import and the actual Component Definition.

kof:
I misinterpreted your suggestion. Looks very promising!

kof:
By reverse order rendering I meant:
```
<SheetB />
<SheetA />
<A><B/></A>
```

So the parent's sheet is rendered lower in the head than the child, and you did that by starting with a negative index.

cvle:
I have been using a similar technique for https://github.com/wikiwi/react-jss-theme (discontinued) and now https://github.com/wikiwi/themr. It works great so far! If you agree I'll finish this PR by adding some tests.

src/index.js
```diff
@@ -86,6 +86,9 @@ const Container = ({children}) => (children || null)
 
 export const jss = createJss(preset())
 
+// global index counter to preserve source order.
+const indexCounter = -10000;
```
nathanmarks: this needs to be `let`, can't `++` a `const`.

cvle: Yes, correct.

kof:
@nathanmarks found one potential case where it will not work: 

```
// Button.js
const Button = (props) => (
  <button>
    {clone(props.icon)}
  </button>
)

// something.js
import Icon from 'icon';
import Button from './Button'

<Button icon={<Icon />} />
```

cvle:
@kof Can you explain your case more in detail?

kof:
So the case is where a generic component accepts a react element, clones that element and tries to override some rule props using an own rule over class name.

Now the specificity of the  rule from Button component depends on whether  the "something.js" module has imported "Icon" first and "Button" second or the other way around. If Icon is imported first and button second, the sheets order will be <button sheet><icon sheet>, so the icon sheet will have a stronger selector.

kof:
I would still go for your solution because I don't know anything better and it will work in the most cases.

kof:
Maybe defining the index inside of the constructor is lightly better, because then it depends on element creation time, not on require() time.

kof:
Lets fix this, I think its even worth an article about specificity issues.

cvle:
Alright, I'll update this PR by tomorrow.

cvle:
Ready to be reviewed

src/index.js
```diff
@@ -86,6 +86,9 @@ const Container = ({children}) => (children || null)
 
 export const jss = createJss(preset())
 
+// global index counter to preserve source order.
+let indexCounter = -10000
```
kof: maybe one more zero, just to be 100% sure?)))

src/index.test.js
```diff
@@ -78,6 +78,70 @@ describe('react-jss', () => {
     })
   })
 
+  describe('.injectSheet() preserving source order', () => {
+    let WrappedComponentA
+    let WrappedComponentB
+    let WrappedComponentC
+
+    before(() => {
+      sheets.reset()
```
kof: why do you need this if you do the reset in afterEach anyways?

kof: Also I would prefer a new instance rather than reseting.

kof:
I guess we would have much nicer tests if we would use enzype + snapshots ...

src/index.js
cvle: Sure.

src/index.test.js
cvle: I do a `sheets.reset()` before, because the other tests running before are `leaky`. I probably should reset the other tests properly instead.

It's not optimal, but I was trying to avoid putting another component in the test (SheetsRegistryProvider) so the test is more isolated. Should we use the SheetsRegistryProvider instead? Or maybe use enzyme and provide the context directly.

cvle:
Snapshot means using `jest` right? I'm not familiar with `jest` sounds like a major change.

src/index.test.js
kof: I think we should create new Jss and injectSheet in beforeEach on a very top. So that there is no need for reseting. I didn't know other tests leak.

kof:
Yeah its a bit a bigger change. Also we would loose testing in real browser, which might be ok for this project, because all DOM stuff happens in the core anyways.

src/index.test.js
cvle: Ah yes, I overlooked that we have a `createInjectSheet` method for this purpose.

kof: I would do this for all tests here, not just for the new once.

cvle: I just realized that the sheet registry is hard coded in `jss` (since v6.0.0?). So I cannot change it. Considering this I would keep using `sheets.reset()`.

I prefer not to add it to the very top and have it explicitly done in the `describe` blocks. This way it is easier to understand (all affecting code is defined locally), especially for new contributors.

kof: what about SheetsRegistryProvider

kof: so you can create a registry at a very top and wrap your component with SheetsRegistryProvider

cvle: > I was trying to avoid putting another component in the test (SheetsRegistryProvider) so the test is more isolated.

Yes, this would be the simplest way, but less isolated. What's your concern about resetting the sheets? Another way is to discard the module cache and require a new jss instance.

kof: >  What's your concern about resetting the sheets? 

Just not a very nice pattern. But if everything else is too complicated I am ok with .reset for now.

cvle: Would you feel more confortable with resetting and reloading modules instead of only resetting the sheet registry? This adds around 20-30 lines of code. I read that facebook is using this pattern with jest (https://facebook.github.io/jest/docs/api.html#jestresetmodules) too.

kof: Would you?))) Also are you migrating to jest right now?

cvle: I'm not, but I used this technique for testing once.

cvle: Spontaneously I would say I feel more comfortable resetting all modules to have a fresh start, as we have global state in jss and react-jss. I'll push a revision with resetting the modules.

kof: ok

kof:
Will merge it for now, though I think this approach is a bit verbose)

cvle:
A bit yes, maybe you can find ways to make it less verbose.

PR #66, opened 2016-12-28
Add code coverage

cvle:


kof:
Can we somehow indicate the coverage on the readme with a badge?

kof:
Initially I was thinking to use codecov.io because it gives you a badge and also shows coverage stats on each PR…

cvle:
I personally use coveralls.io. I think it offers the same features but I haven't been using the PR feature yet.

cvle:
With this PR we have the coverage information after running test, we then need to send it to the codecov service. Normally I send coverage information using travis, but we don't have travis setup for the plugins. If we want PRs to show coverage stats, we will need travis to setup here too.

kof:
We need to setup travis anyways. So you think we still can use codecov or coveralls with this setup? If yes then lets merge this first and do those things later.

cvle:
Exactly. With this PR all you need to do is to run the test and use the CLI from codecov / coveralls.

PR #68, opened 2017-01-30
Pass classes directly to component

### LINKED ISSUE #67: Pass only classes, but not whole sheet?
Currently I am receiving the whole `sheet` object into the component:

```js
const Button = ({sheet: {classes}}) => (
  <button className={classes.button}>Click me</button>
);
```

However, I am using only `classes` property. So, it could be much shorter

```js
const Button = ({classes}) => (
  <button className={classes.button}>Click me</button>
);
```

Not a big deal, but it is very overwhelming to write `{sheet: {classes}}` for *every* component. What else can I do with `sheet` object within the component? I don't see any, so I'd like to write less.

just-boris:
without the rest of the sheet object
closes #67

----

* [x] think, what happens when `this.sheet` is not defined
* [x] add tests on new functionality

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -80,7 +80,7 @@ function wrap(jss, WrappedComponent, stylesOrSheet, options = {}) {
     }
 
     render() {
-      return <WrappedComponent sheet={this.sheet} {...this.props} />
+      return <WrappedComponent sheet={this.sheet} classes={this.sheet.classes} {...this.props} />
     }
   }
 }
--- a/src/index.test.js
+++ b/src/index.test.js
@@ -105,6 +105,32 @@ describe('react-jss', () => {
     })
   })
 
+  describe('.injectSheet() classes prop', () => {
+    let passedClasses
+    let WrappedComponent
+
+    beforeEach(() => {
+      const Component = ({classes}) => {
+        passedClasses = classes
+        return null
+      }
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -80,7 +80,7 @@ function wrap(jss, WrappedComponent, stylesOrSheet, options = {}) {
     }
 
     render() {
-      return <WrappedComponent sheet={this.sheet} {...this.props} />
+      return <WrappedComponent sheet={this.sheet} classes={this.sheet.classes} {...this.props} />
```
kof: we should pass it only if it isn't already passed by the parent component, like this

classes={this.props.classes || this.sheet.classes}

kof:
`think, what happens when this.sheet is not defined` can't be undefined

src/index.js
giuseppeg: `{...this.props}` should override `classes={this.sheet.classes}` I believe

kof: ah right, sorry!

kof:
so we just need to add a test

just-boris:
Yes, will do it at evening.
Currently, I am full with my work...

just-boris:
Done.

kof:
merged

kof:
thanks!

PR #83, opened 2017-03-31
don't ignore static styles when compose dynamic/static - fix #82

### LINKED ISSUE #82: I can only get static or dynamic sheets but not both
[This commit](https://github.com/cssinjs/react-jss/commit/895e5feb766c33d3f2f27d13b4b5ecc1b3d3200f#diff-1fdf421c05c1140f6d71444ea2b27638R124) seems to break the whole way the classes are passed to the component as we can get either the dynamic or the static ones. I would assume to get both. 

There was also a compose function added but this one dont copy over classes that are in the static but not in the dynamic sheet.

eskimoblood:


kof:
merged, thanks!

PR #95, opened 2017-06-11
feature: initial theming

iamstarkov:

PR #96, opened 2017-06-13
feature: advanced theming [WIP, dont merge] 

iamstarkov:
published as `@iamstarkov/react-jss-theming-plus@6.1.1`:

* old school `injectSheet` works as expected
* initial theming function from #95 works perfectly fine
* advanced theming doesnt work, but at least it doesnt break shit
* source code is in here
* playground demo to take a look is here  https://www.webpackbin.com/bins/-KmYmlKB42LBQaxDJsVK

i couldnt to make it work locally with npm@5 and `npm link` and therefore i publishe my local build to `@iamstarkov/react-jss-theming-plus`. In order to do the same you need:
* change `pkg.name` from `react-jss` to `@${your-npm-username}/react-jss-${suffix}`
* run `npm publish`

@kof i would need your help to figure out how to integrate `themeListener` in the `createHoc`. pls, take a look


iamstarkov:
well, i got it working. though themeListener needs some adjustments

iamstarkov:
related https://github.com/iamstarkov/theming/pull/3

src/createHoc.js
```diff
@@ -110,7 +111,12 @@ export default (jss, InnerComponent, stylesOrSheet, options = {}) => {
       }
     }
 
-    componentWillUpdate() {
+    componentWillUpdate(nextProps, nextState) {
+      if (isThemingEnabled && nextState.theme && JSON.stringify(this.state.theme) !== JSON.stringify(nextState.theme)) {
```
kof: Thats too expensive, we should compare by ref and treat it as immutable. Thats what react and other do in such cases.

For a perfect dx we could add warnings in the dev mode  to warn user if he mutates a theme. Actually this should be part of theming.

kof: You also need  a flag which indicates if styles have been returned by a function or they are statically defined. In the later case, you don't need to do anything here.

iamstarkov: i know that its too expensive, i will refactor it later. that said its not my main concern rn

iamstarkov: > You also need a flag which indicates if styles have been returned by a function or they are statically defined. In the later case, you don't need to do anything here.

i believe its `isThemingEnabled`

iamstarkov:
me and @kof (mainly @kof) managed to get it to work
uwa4rm2fuw

kof:
Changed the target to the next branch, moved all recent changes from master to the next branch.

iamstarkov:
> Changed the target to the next branch, moved all recent changes from master to the next branch.

@kof why?

kof:
because #98 - people read documentation and wondering why it doesn't work, lets merge to master when we release.

iamstarkov:
maybe better option is to release ThemeProvider and withTheme?

kof:
How long is it until we can merge this PR?

kof:
I thought its just a couple of days left, so its better to wait

iamstarkov:
why should we wait? this is advanced theming, people still can use themeprovider and withtheme

kof:
Because after this is merge, they don't need to use withTheme by default for regular components with styles. I wanted to make a big announcement once it is ready, changing this after that is not very nice.

iamstarkov:
okay, i got you

iamstarkov:
i will need your help, btw with `hot-reload` and `StyleSheet arg`

jcheroske:
I would really like to give this a test drive. Is the API stable at this point? Are there any docs I could look at?

kof:
take a look at "next" branch

kof:
The only change will come is that you don't need to use "withTheme()" for components you want to style with JSS.

jcheroske:
So, is the major change that, instead of just being able to pass and object to `injectSheet`, we'll now be able to pass a function that will get the theme (and hopefully the props too)? I'll look at the code. Thanks for all the hard work.

kof:
> (and hopefully the props too)

no props, theme only, thats important for caching.

jcheroske:
Gotcha. Not a problem at all. I'm going to be replacing our theming solution with this as soon as it's ready. Do you know if `ThemeProvider`s can be nested?

kof:
> Do you know if ThemeProviders can be nested?

They can!

kof:
Regarding hot reload lets ask the only source of truth @gaearon. Whats the current status, which of the implementations should we look at? I am a bit lost because I stopped using it at some point.

iamstarkov:
@kof here is an idea

iamstarkov:
as far as keeping static sheets in the JSS hoc was a good idea, lets keep it

iamstarkov:
then, we need to share static Sheets, so i introduced `SharedStaticSheets` per each JSS hoc

iamstarkov:
```js
class SharedStaticSheets {
  constructor() {
    this.sheets = []
  }

  // remove all sheets that are not used any more
  clean() {
    this.sheets = this.sheets.filter(item => item.count !== 0);
  }
  // add sheet if its not already there, return sheet from this.sheets
  add(sheet) {
    let index = this.sheets.findIndex(item => item.hash === hashify(sheet))
    if (index === -1) {
      this.sheets = this.sheets.concat({ sheet, count: 0, hash: hashify(sheet) })
      index = this.sheets.length - 1
    }
    return this.sheets[index].sheet;
  }
  // attach sheet if its already not, and increment count
  attach(sheet) {
    let index = this.sheets.findIndex(item => item.hash === hashify(sheet))
    if (index === -1) {
      throw new Error('[sharedStaticSheets], you can attach non-added sheet: ' + sheet)
    }
    // * if counter is 0, attach
    if (this.sheets[index].count === 0) {
      this.sheets[index].sheet.attach()
    }
    // * inc count
    this.sheets[index].count = this.sheets[index].count + 1
  }
  // decrement count of the sheet, detach it if count is zero, clean sheets
  detach(sheet) {
    let index = this.sheets.findIndex(item => item.hash === hashify(sheet))
    if (index === -1) {
      throw new Error('[sharedStaticSheets], you can detach non-added sheet: ' + sheet)
    }

    // * if its there, dec count
    this.sheets[index].count = this.sheets[index].count - 1

    // * if counter is 0, detach
    if (this.sheets[index].count === 0) {
      this.sheets[index].sheet.detach()
    }
    this.clean();
  }
}
```

iamstarkov:
this way, static sheets are shared among mounted components

iamstarkov:
also the same idea can be used to dedupe dynamic sheetshttps://github.com/cssinjs/jss/issues/501

iamstarkov:
one caveat is in order to dedupe we need an heuristic to distiungish Sheets. You cant use reference based check, because static sheet in each mounted component is new instance of the sheet, so check wont work.

iamstarkov:
i came up with working proof of concept of `const hashify = sheet => createHash(JSON.stringify(sheet.rules.raw))`, it might be not performant enough, but it works

iamstarkov:
second problem is that i couldnt find a way to not duplicated sheets into sheetsRegistryProvider

iamstarkov:
if we could delegate abilities of mine `SharedStaticSheets` to `sheetsRegistryProvider` then it will be easier. and then mental map will be easier as well:

if Sheet is in the `sheetsRegistryProvider`, then it will end up in the DOM and will be used in SSR

iamstarkov:
so here are my results on advanced theming for react-jss so far. What do you think?
/cc @kof

iamstarkov:
regarding hot-reload, if example will be added to next.js https://github.com/cssinjs/jss/issues/457, then it will be easier to verify that it works

iamstarkov:
1) sheets reusability by props and theme:
a) if theme is shallow equal to next theme, then staticSheet is the same.
b) if theme is changed then extractDynamicStyles from staticSheet.
c) if nextStaticStyles are shallow equal to prevStaticStyles, then reuse static sheet, otherwise we generate new staticSheet
d) remove old dynamicSheet, create new dynamicStyles'
/cc @kof

src/createHoc.js
```diff
@@ -1,13 +1,85 @@
...
+add:
+  * if its not already there, add it with count: 0,
+  * if counter is 0, attach
+  * inc count
+remove:
+  * if its not already there, throws
+  * if its there, dec count
+  * if counter is 0, detach
+*/
+class SharedStaticSheets {
```
iamstarkov: yo, @kof

src/createHoc.js
```diff
@@ -1,7 +1,8 @@
 import React, {Component} from 'react'
 import {object, instanceOf} from 'prop-types'
-import jss, {SheetsRegistry, getDynamicStyles} from 'jss'
+import {SheetsRegistry, getDynamicStyles} from 'jss'
 import { themeListener } from '@iamstarkov/theming-w-listener'
+import jss from './jss'
```
iamstarkov: if we use local jss instance, then we should use local `SheetsRegistry` and `getDynamicStyles`, shouldnt we?

```js
import jss, {SheetsRegistry, getDynamicStyles} from './jss'
```

kof: we can, yes

kof: its same shit though.

kof: done

iamstarkov:
finished all the tests for theming

iamstarkov:
before merging this:
* jss needs a new feature release w/ sheetsManager
* theming needs a patch release with fix for themes' references issue
* theming needs a feature release theme-listener
* react-jss needs a pre-release to test it out with next.js (ssr and hot-reload)

iamstarkov:
that means, we almost get it

iamstarkov:
seems like everything is cool

iamstarkov:
but, i cant make pre-release

iamstarkov:
@kof can you add me a collaborator to this project on npm?

kof:
done

kof:
merged :)

PR #107, opened 2017-07-06
Fixed problem with nondeterministic styles resolution during SSR

a-ignatov-parc:


src/index.test.js
```diff
@@ -450,6 +451,127 @@ describe('react-jss', () => {
...
       `)
     })
+
+    it('should be idempotent', () => {
+      const mapToProp = name => props => props[name]
+
+      const Component = injectSheet({
+        button: {
+          color: mapToProp('idle'),
+          '&:hover': {color: mapToProp('hover')},
```
kof: why so much complexity? Wouldn't the issue be reproduced just by using one property?

src/index.test.js
```diff
@@ -450,6 +451,127 @@ describe('react-jss', () => {
...
+          '&:active': {color: mapToProp('active')},
+          '&:visited': {color: mapToProp('visited')}
+        }
+      })()
+
+      const customSheets1 = new SheetsRegistry()
+      const customSheets2 = new SheetsRegistry()
+      const customSheets3 = new SheetsRegistry()
+
+      ReactDOMServer.renderToString(
```
kof: same here, why 3 renderToString calls? Can it be 1? 2?

src/index.test.js
```diff
@@ -450,6 +451,127 @@ describe('react-jss', () => {
...
+            idle="#000"
+            hover="#333"
+            active="#666"
+            visited="#ccc"
+          />
+        </JssProvider>
+      )
+
+      // Removing appended styles after `ReactDOMServer.renderToString`
+      Array
```
kof: renderToString doesn't append any styles, why is this really needed?

a-ignatov-parc: `renderToString` doesn't but `react-jss` is applying styles

src/index.test.js
a-ignatov-parc: Good catch

src/index.test.js
a-ignatov-parc: I'll check it

src/index.test.js
kof: we normally unmount after each test so they are removed, as you are using DOM renderer it shouln't be rendering styles at all, but we are in the browser here, so it does. Best way to solve this is to pass virtual: true manually as Jss option.

a-ignatov-parc: Sounds good 👍

a-ignatov-parc:
@kof fixed all comments

src/createHoc.js
```diff
@@ -19,17 +19,9 @@ const inc = sheet => ++sheet[refNs]
...
-  let staticSheet = null
-  let dynamicStyles
-
-  // Accept StyleSheet instance.
-  if (stylesOrSheet && typeof stylesOrSheet.attach === 'function') {
-    staticSheet = stylesOrSheet
-    styles = null
-  }
-
+  const stylesIsInstance = stylesOrSheet && typeof stylesOrSheet.attach === 'function'
```
kof: I like your semantics, nice.

kof:
Will continue CR later today, stay tuned.

src/createHoc.js
a-ignatov-parc: Thanks )

a-ignatov-parc:
Sure. Great :+1:

src/index.test.js
```diff
@@ -450,6 +451,73 @@ describe('react-jss', () => {
...
+        </JssProvider>
+      )
+
+      const result1 = customSheets1.toString()
+      const result2 = customSheets2.toString()
+
+      expect(result1).to.equal(result2)
+    })
+
+    it('should render deterministically on server and client', () => {
```
kof: This test did pass before your changes as well. What is special about it? Do we really need it?

src/index.test.js
```diff
@@ -3,6 +3,7 @@
 import expect from 'expect.js'
 import React, {PureComponent} from 'react'
 import {render, unmountComponentAtNode, findDOMNode} from 'react-dom'
+import ReactDOMServer from 'react-dom/server';
```
kof: please run npm run lint

src/index.test.js
```diff
@@ -450,6 +451,73 @@ describe('react-jss', () => {
         }
       `)
     })
+
+    it('should be idempotent', () => {
```
kof: this test is quite interesting, it really nails the problem

src/createHoc.js
kof: The problem with the current implementation is that me and @iamstarkov are in the middle of a big refactoring of this modules. We are adding theming support right to the HOC. So all the code changes will be rewritten anyways. I am mostly interested in tests.

kof: I can still merge it though so that commit history reflects your effort.

src/index.test.js
a-ignatov-parc: I think we should keep it because we really need to be sure that server and client rendering are always make same result ;)

src/index.test.js
a-ignatov-parc: Yes that was the hardest part to pin point the minimal test case )

src/createHoc.js
a-ignatov-parc: It's you to decide on what to do with this PR. I'm only interested in final result ;)

a-ignatov-parc: Here is https://github.com/cssinjs/react-jss/pull/108 with standalone test cases if you decide to merge only tests

kof:
You were right, since the classes generation algorithm change, we can't keep rendered sheets even if they don't really change, they need to be regenerated.

kof:
we got this covered in the next branch now.

a-ignatov-parc:
@kof great! Looking forward for a new release ;)

PR #110, opened 2017-07-12
add an option for "classesNamespace" to alias the "classes" prop

rob2d:
This would make it easier to create HOC's which contain their own JSS with an inner component that has a different style sheet to inject without conflict. 

It also addresses global prop namespace resolution (which is one concept that was about CSS) when introducing ReactJSS into other projects that may want to use that for any other purpose (e.g. any educational app with the plural noun `classes`).

(Default behavior is still retained where `classes` prop is assigned as `classes`)

Super simple (arbitrary) example:

```
injectSheet( styleSheet, { classesAlias : `otherClasses` })
( ({ otherClasses })=>(<div className={otherClasses.container})>content</div>) )
```


kof:
Why not to pass a different name than classes from the HOC? 

```js
<child myClasses={{}}>
```

rob2d:
If using an HOC around an HOC, regardless of how many times `injectSheet` is processed in scope, it only funnels one set of `props`. So (unless I'm mistaken), it doesn't seem technically possible to reference those props, while still having the `<child>` in that example also as wrapped by `injectSheet`. The problem with this is we end up needing to create a 3rd component to separate the abstraction between the two, which is kind of overkill/anti-pattern for HOCs.

kof:
Can you show a very simple but realistic example?

rob2d:
No problem. Here is the elaborated idea showing the limitation until I go adapt a UI library issue I'm having now :

```
const hocStyle =
{
    container :
    {
        width : '100%',
        height: '100%',
        //... arbitrary style here ...
    }
};

const aStyledHOC = ((WrappedComponent)=>(injectSheet(hocStyle)(props)=>
{
    const { classes, ...propsWithoutClasses } = props;

    return (
        <div className={props.classes.container}>
            <child {...propsWithoutClasses} />
        </div>
    );
}));

const styleSheet =
{
    myOwnStyle : 
    {
        // nothing here can ever reach the child component (!)
    }
};

const anotherComponent = aStyledHOC(injectSheet(styleSheet)({ classes }=>
(
    <div className={classes.myOwnStyle}>
        My classes are not here
    </div>
);
```

kof:
> injectSheet(hocStyle)(props)

This makes no sense. Second invocation should receive a component.

rob2d:
typo; fixed.

rob2d:
```
const DEFAULT_FADETIME = 1000;
const faderStylesheet =
{
    container :
    {
        transition : ({ fadeTime = DEFAULT_FADETIME })=> (
            `visibility ${fadeTime}, opacity ${fadeTime}`
        ),
        visibility : ({ isShown })=> (
            isShown?'hidden':'visible'
        ),
        opacity : ({ isShown })=>(isShown ? 0 : 1)
    }
};

const componentFader = ((WrappedComponent)=>(injectSheet(faderStylesheet)((props)=>
{
    const { classes, ...propsWithoutClasses } = props;

    return (
        <div className={props.classes.container}>
            <WrappedComponent {...propsWithoutClasses} faderClasses={classes} />
        </div>
    );
}));

const styleSheet =
{
    container : 
    {
        backgroundColor : '#FF0000',
        color           : '#FFFFFF',
        padding         : '16px'
    }
};

const RedJSSComponent = componentFader(injectSheet(styleSheet)({ classes }=>
(
    <div className={classes.container}>
        Hello world. I am a red JSS Styled Component which inherited from
        a JSS HOC.
    </div>
));

class Example extends PureComponent
{
    constructor(props)
    {
        super(props);
        this.state = { isShown : false };
    }
    toggleIsShown = ()=>
    {
        this.setState({ isShown : !this.state.isShown });
    }
    render ()
    {
        <div onClick={this.toggleIsShown}>
            Click me to show a Red JSS Component
            <RedJSSComponent isShown={this.state.isShown}>
        </div>
    }
};
```

The example above does not currently work as expected because you cannot have an HOC which relies on `injectSheet` wrapping another JSS component, so the red div will never fade in upon clicking despite being generated from `injectSheet` with a stylesheet.

With a `classNamespace` param in the `options`, its possible to create HOCs that don't interfere and can rely on JSS intuitively and without overengineering:
```
const componentFader = ((WrappedComponent)=>(injectSheet(faderStylesheet, { classNamespace : 'faderClasses' })((props)=>
(
        <div className={props.faderClasses.container}>
            <WrappedComponent {...props} />
        </div>
)
));
```

kof:
I fail to see the problem https://www.webpackbin.com/bins/-KovsqR2hJjhJu8aXJC7

iamstarkov:
i dont understand the use case. @rob2d, can you elaborate a bit more? (sorry, im a bit tired)

rob2d:
Sorry, that example does seem to be working. And feeling very dumb here 😢 I had originally had a mistake in the logic which I corrected while posting here. Had written that in a rush since I was at work.

I edited your WebpackBin and things work 100% here:
https://www.webpackbin.com/bins/-KoyeW4zkMotyQIecLu1

So in this case, I will mark this as closed. I think it is a legitimate idea to be able to pass that as an option to avoid collision, but its definitely opinionated and unnecessary. As @kof mentioned, if you skip over props and funnel the HOC classes prop into another through the child. 

I appreciate both of your time and patience with that knuckleheadedness there as well as the work on this.

(Also thanks for the awesome webpackbin resource; was not aware. I will use that going forward 😄)

Issue #121, opened 2017-07-18
JssProvider breaks injectSheet with enabled theming


HenriBeck:
Apparently, the theming isn't working with nested components that require theming.
It seems that the nested component will get the sheet and classes from the upper component.

I may have done something wrong but when I log some values, it seems that [this line](https://github.com/cssinjs/react-jss/blob/master/src/createHoc.js#L80) returns the sheet from the upper component which leads to not call the actual style creator function.

kof:
I assume you did something wrong. Please provide code for reproduction.

HenriBeck:
https://github.com/HenriBeck/materialize-react/tree/update-theming

Install packages and start storybook.
The upper component is the Background component.
The Ripple element should be working as the ripple does not depend on the theme.
The Button component, on the other hand, does not work and it will log out the classes for the Background component when rendered.

kof:
> I can provide a more organized example by tomorrow.

That would be nice.

HenriBeck:
https://github.com/HenriBeck/react-jss-theming-example

As you can see in the console, the button component does not log the correct object of classes and instead logs the classes of the Background component

kof:
It sounds like you are passing classes down from your parrent and overriding JSS HOC classes by it.

kof:
This sounds pretty much unrelated to theming.

HenriBeck:
No, I'm not passing any classes down.
If I go into the react-jss package and log the [staticStylesheet](https://github.com/cssinjs/react-jss/blob/master/src/createHoc.js#L80) there, once it returns undefined for the Background component and for the button component it returns the sheet from the Background component. That's why I end up with different classes in the Button component.

iamstarkov:
its odd, im investigating

iamstarkov:
and yes, you are not passing classes from one component to another. its good

iamstarkov:
you had some obvious errors, i fixed it in https://github.com/HenriBeck/react-jss-theming-example/pull/1

iamstarkov:
TLDR: 
* you used composes wrong
* you used destructuring wrong in your styles creator functions

iamstarkov:
and you used `JssProvider` when you didnt need

iamstarkov:
nevertheless, problem is actually in JssProvider

iamstarkov:
root problem is in JssProvider https://github.com/iamstarkov/react-jss-theming-example/pull/1

iamstarkov:
[...]

PR #122, opened 2017-07-18
add failing test for #121

iamstarkov:


kof:
Fixed, are we good to go?

iamstarkov:
i am concerned that right now tests are covering only SSR

iamstarkov:
while error happened without it

iamstarkov:
added tests for that

src/createHoc.js
```diff
@@ -73,7 +73,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
     }
 
     get manager() {
-      return this.context[ns.sheetsManager] || manager
+      if (providerId && this.context[ns.providerId] !== providerId) {
+        manager = new SheetsManager()
+      }
+      providerId = this.context[ns.providerId]
+
+      return manager
```
iamstarkov: can you describe in plain text what is happening here?

iamstarkov: nvm

kof:
> i am concerned that right now tests are covering only SSR

JssProvider was involved, it is most of the time used for SSR, also it has reproduced the issue, so I wouldn't know what to cover otherwise.

iamstarkov:
@kof original issue was reproducible with `JssProvider` but without ssr, so i added tests for that

kof:
It was reproducible with both, renderToString and render, so it doesn't matter. The nature of the issue was independent of environment. I removed those other tests because after fixing the issue covered by ssr test other tests were fixed too, so I though it doesn't make sense to have tests where it is not 100% clear why they are needed.

iamstarkov:
my idea behind additional test is that issue is reproducible without SSR, so i thought it makes sense to have additional test wo/ ssr

src/JssProvider.js
```diff
@@ -27,7 +26,7 @@ export default class JssProvider extends Component {
       [ns.sheetOptions]: {
         generateClassName: createGenerateClassName()
       },
-      [ns.sheetsManager]: new SheetsManager(),
```
oliviertassinari: I wish we have kept that feature for https://github.com/callemall/material-ui/pull/7461.

src/createHoc.js
```diff
@@ -73,7 +73,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
     }
 
     get manager() {
-      return this.context[ns.sheetsManager] || manager
+      if (providerId && this.context[ns.providerId] !== providerId) {
+        manager = new SheetsManager()
```
oliviertassinari: I think that it's breaking concurrency in SSR env.

iamstarkov: what do you mean?

iamstarkov:
@oliviertassinari i cant find you comment

> I wish we have kept that feature for https://github.com/callemall/material-ui/pull/7461

oliviertassinari:
I have removed this comment. I have added that feature to the MuiThemeProvider instead of the JssProvider.

iamstarkov:
i mean we fixed a bug, why would you rely on it?

src/createHoc.js
oliviertassinari: Let's say you are using https://formidable.com/blog/2017/introducing-rapscallion/, you are going to leak information between requests. That's bad for the UI and for security.

kof: need to add a test for this

PR #123, opened 2017-07-21
Fix hmr

iamstarkov:
so, the problem is in `jss.SheetManager.unmanage()` it will throw if there is no sheet to unmanage.  HMR updates remounts the component and SheetManager is being resetted, so it will throw, but fix is easy, just dont throw if its HMR update


iamstarkov:
this branch is published as `@iamstarkov/react-jss@7.0.2` if you want to give it a try

iamstarkov:
@kof i need your review

src/createHoc.js
```diff
@@ -142,10 +142,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
     }
 
     componentWillUpdate(nextProps, nextState) {
-      if (isThemingEnabled && this.state.theme !== nextState.theme) {
+      const isThemeUpdate = isThemingEnabled && this.state.theme !== nextState.theme
+      const isHMRupdate = process.env.NODE_ENV !== 'production' && this.manager.keys.length === 0
```
kof: we are using strict camel case, meaning even if its abbreviation, still isHmrUpdate

kof: manager.keys is somewhat private

iamstarkov: @kof maybe, but that will cause the change in the jss

iamstarkov: renamed

kof: yes a minor release.

iamstarkov:
@kof @izorg ready to re-review

src/createHoc.js
iamstarkov: ok, hold on

iamstarkov: nay

iamstarkov: if we would do so, then next release of react-jss will behighly dependent on next release of jss, otherwise it will break

iamstarkov: so, even if we get size support into sheetsmanager, we would still need to maintain `manager.keys.length`

iamstarkov: but, i'll submit  pr

kof: true, for react-jss it would be a major release

iamstarkov: @kof see https://github.com/cssinjs/jss/pull/544

patrickml:
@iamstarkov how long does it normally take before a PR like this is merged and a new version is published?

kof:
@patrickml there is no average time span, every task is different. It is waiting not because we do nothing.

kof:
We communicate  in dm sometimes. We need a test for hmr fix which doesn't use internals directly and we are looking for a solution.

patrickml:
is there a reason why you couldn't use `module.hot.accept('./hmr-style-test', () => {})` and use `fs` to update the file to trigger a change? I know it doesn't sound ideal to run an hmr server for the test, but at least it'd be a true integration test and should show breakages with future hmr updates>

kof:
Do you know how to integrate it into the tests?

patrickml:
No, it was just an idea, but im looking at the react-hot-loader package and maybe the answer lies within there tests? https://github.com/gaearon/react-hot-loader/blob/master/test/patch/patch.dev.js#L57

They seem to have a `register` function that i'm assuming makes the change

```js
      const RHL = global.__REACT_HOT_LOADER__;
      RHL.register(A1, 'a', 'test.js');
      const A = <A1 />.type;
      expect(A).toNotBe(A1);
      expect(A).toBeA('function');
      expect(<A />.type).toBe(A);

      RHL.register(A2, 'a', 'test.js');
      expect(<A1 />.type).toBe(A);
      expect(<A2 />.type).toBe(A);
      expect(<A />.type).toBe(A);
```

iamstarkov:
i would like to find and try every possible way before trying to run server and manually changing test files, that would be just a disaster

patrickml:
Well it looks like their test doesn't actually run the server. They are just importing this file https://github.com/gaearon/react-hot-loader/blob/master/src/patch.dev.js to load `__REACT_HOT_LOADER__ ` into the global namespace

iamstarkov:
that might make sense

src/createHoc.js
```diff
@@ -140,12 +148,19 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
...
+     */
     componentWillUpdate(nextProps, nextState) {
-      if (isThemingEnabled && this.state.theme !== nextState.theme) {
+      const isThemeUpdate = isThemingEnabled && this.state.theme !== nextState.theme
+      const isHmrUpdate = process.env.NODE_ENV !== 'production' && this.manager.keys.length === 0
+      if (isThemeUpdate || isHmrUpdate) {
         const newState = this.createState(nextState)
         this.manage(newState)
-        this.manager.unmanage(this.state.theme)
+        if (!isHmrUpdate) this.manager.unmanage(this.state.theme)
```
oliviertassinari: Won't we end up with unremoved sheets in the page?

src/createHoc.js
```diff
@@ -49,6 +49,14 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
 
   const displayName = `Jss(${getDisplayName(InnerComponent)})`
   const noTheme = {}
+
+  /**
+   * manager will be resetted after HMR is triggered,
```
oliviertassinari: I don't think that it's rested, the hooks are executed in a different closure context where the `manager` isn't initialized yet.

src/createHoc.js
kof: We will. The assumption is that it makes no troubles in dev.

kof: which is actually wrong, because if styles use @global, they can cause issues, I assume people are not using that too often though.

oliviertassinari: You can have a look at the fix I have done on Material-UI side, we can avoid that issue.

kof: do you have a link?

oliviertassinari: https://github.com/callemall/material-ui/pull/7601/files

oliviertassinari: Some users had the very same issue. For some reason it's not visible on next.js nor our documentation website, still it can happen. You can attach the closure variable to the instance of the context to dodge that issue.

src/createHoc.js
iamstarkov: it is resetted in the manager getter

iamstarkov: because of different closure contexts

src/createHoc.js
iamstarkov: @oliviertassinari i looked at your fixes and couldnt grasp how to port your fix here. can you help?

oliviertassinari: I would try attaching `manager` to the instance of the component.

iamstarkov: @oliviertassinari unfortunately, that will destroy the reason of manager, because manager is meant to be shared among instances of the same component, so sheets could be shared among components as well thus avoiding duplication along the way

oliviertassinari: It doesn't mean it can't be shared.

Also, it's a limitation of this approach you can't share the same sheet across different components classes (as well as instances) while material-ui design allows it (but it's not important).

felthy:
I’d love to get hot module replacement working for JSS styles in my projects so I’ve had a go at bringing this pull request up to date. I didn’t find a way to contribute directly to this pull request so I opened a new one #266 based on the `fix/hmr` branch. Please let me know if there’s a better way to contribute that wouldn’t involve starting a new pull request 🙂

kof:
@felthy you can now contribute directly from this branch, once you accepted the invitation

felthy:
Thanks for the invite @kof, I’ve merged my updates into this PR now so I’ll close #266. All tests are passing now.

I’ll have a think about [your suggestion](https://github.com/cssinjs/react-jss/pull/266#issuecomment-401736218) of a more general API - I agree that having HMR-specific code in this package isn’t ideal.

kof:
But I am ok with having it in this package, just would be nicer to at least have it in a separate module.

thiphariel:
So, is this PR will be merged ? I use react-hot-loader too and I would like to update my styles with HMR, for now, I use this PR (fix/hmr branch)

HenriBeck:
Can we add some documentation on how to setup HMR before merging? @felthy

felthy:
@thiphariel I’ve been working on a standalone package as per the discussion with @kof above. I’ve still got some work to do on the tests but if anyone would like to try it, you can clone it from [felthy/react-jss-hmr](https://github.com/felthy/react-jss-hmr) and perform a build (`npm run build`). The README could do with a bit more detail as well but hopefully there’s enough info there to get you going.

`react-jss-hmr` doesn’t require any changes to `react-jss` so you don’t need to use this branch with it - I’ve been testing with the current published version, 8.6.1.

@HenriBeck @kof I’ll be interested to get your feedback on this. My hope is that we can move it into the `cssinjs` org if we all agree it’s the best way forward.

felthy:
Update: Tests are working nicely (migrated to `jest`) and the package is available on npm as [react-jss-hmr](https://www.npmjs.com/package/react-jss-hmr).

thiphariel:
@felthy Okay ! I gonna give it a try right now ! Thanks. I will leave some feedback in few days after testing :)

HenriBeck:
@felthy Yes, moving that too the `cssinjs` org would be great. 
Do the users still need to wrap every `injectSheet` call with the `hot` HOC?
Because that's something I would like to avoid.

felthy:
@HenriBeck No, you only need to use the `hot` HOC once, to wrap your top level component (e.g. `App` in the create-react-app starter project). It doesn’t matter whether that component uses `injectSheet` or not either.

kof:
@felthy since we are moving to a monorepo, I think a good place would be in jss/pacakges

kof:
@iamstarkov @AleshaOleg you may want to check it out https://github.com/felthy/react-jss-hmr

felthy:
OK @kof, I’ll start a PR to add it to the jss monorepo

felthy:
Also FYI we had some back and forth on felthy/react-jss-hmr#1 and as a result @thiphariel got it working with webpack 4 and webpack-dev-server 3. 

I’ve been mostly testing with webpack 3.12.0 but have also confirmed that it works with webpack 4.15.1.

kof:
@HenriBeck yes, I think a good rule would be everything that depends on jss, should be in that monorepo.

HenriBeck:
Okay, then it makes sense to add the `react-jss-hmr` eventually to the monorepo.

Can we move the repo just under the `cssinjs` org until we can merge it into the monorepo?

kof:
sure

HenriBeck:
@felthy Can you move the repo to the `cssinjs` org then?

felthy:
Sure, I’ll do that first then I’ll try bringing it into the monorepo.

felthy:
Here it is: [cssinjs/react-jss-hmr](https://github.com/cssinjs/react-jss-hmr) 😄

HenriBeck:
Great Job 👍. So we can close this pr now?

felthy:
I think so - I might open a new one to mention the new package in this package’s readme though.

kof:
@felthy https://github.com/cssinjs/react-jss-hmr/blob/8bd0b0ab06b57d9c45ca5b2785cae68109d3405b/src/injectSheet.js#L13

Q: why do we have this for production? This shouldn't be included at all. I would rather add

felthy:
Who should I invite to become maintainers on [npm](https://www.npmjs.com/package/react-jss-hmr)? @kof? @iamstarkov?

kof:
I think its ok if you keep control over the publishing, in case you are not available add me

felthy:
Yes a warning is a better idea @kof. I’ll add one.

kof:
Reading through the code, another problematic thing is: it is using private methods and properties a lot, which will lead to broken hmr if react-jss gets refactoring.

kof:
Now knowing it is not using a public interface, but many private things, it should be part of the react-jss package, so that when changing react-jss one can see failing tests in hmr ...

felthy:
You mean `this.manager` and `this.createState()`? Any others?

kof:
and `this.state.theme` and all the react hooks

kof:
I think you could achieve the same using a HOC, you can use react-hooks there, manager and theme comes from context

kof:
We could pass manager over props and make react-jss hoc take it if provided

kof:
So, we can add to react-jss what is needed and make it a public interface just for hrm or we can make hmr part of react-jss ... but we can't make a module that uses private stuff and publish it as a separate package, that would be hard to maintain.

felthy:
The manager won’t be passed down from the HMR HOC to the react-jss one, it goes in the other direction. But in general I think you’re right - I didn’t think most of these things would have been considered private, but that being the case, I think the idea of a separate package will not work. A public interface that exposes enough to make this work will be exposing too much, and there wouldn’t be any other use case for that interface. Tomorrow I will look at bringing this PR in line with the improvements I have in the new package (such as not calling `setState()` from within `componentWillUpdate()`; and not using the stack trace trick, and the better tests).

kof:
We can still have it in a separate module though, in react-jss repo

HenriBeck:
Okay, can we have the discussion in the new repo and close this?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kof:
Its ok to continue here, if we are going to have a module in this repo

HenriBeck:
Ohh yeah, didn't there were new comments.

Can we create a new clean branch and pr then?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

felthy:
I don’t think we should create a new PR, the code in this one does currently work and this page has pretty good SEO by now! 😉 
Really we’re still just talking about improvements to the same new feature.

felthy:
@kof just to clarify, you are ok with using the same technique as react-jss-hmr (monkey-patching, and relying on internals), within a new react-jss module, correct? 

Otherwise, I’ve been trying to think up a way to make the existing createHoc more amenable to HMR but it seems likely to require significant restructuring, which is probably best avoided.

kof:
@felthy I missed your last message, so our status is to use that separate package, right?

felthy:
Hi @kof, yes fine to ignore that message now. Given that the React HMR landscape looks like it will change a lot soon, I think the separate package, ugly as it is, is appropriate. My [last comment over there](https://github.com/cssinjs/react-jss-hmr/issues/2#issuecomment-405833876) is my current plan, but unfortunately I've been busy and haven't made any progress writing that doco yet.

PR #124, opened 2017-07-21
Another failing test case for SSR rendering

a-ignatov-parc:
This issues appeared in `react-jss@7.0.2`. 

If you'll try to render same component with and without `JssProvider` you'll end up with different counters in class names.


kof:
Is it real life case? How did it happen?

a-ignatov-parc:
Yes it's from real app.

Here is a piece of my SSR code:

```jsx
import React from 'react';
import ReactDOMServer from 'react-dom/server';
import {html} from 'common-tags';
import {Helmet} from 'react-helmet';
import {StaticRouter} from 'react-router';
import {Resolver} from 'react-resolver';
import {JssProvider, SheetsRegistry} from 'react-jss';

import Site from 'site';

export default function AppMiddleware(ctx) {
  return Resolver
    .resolve(() => (
      <StaticRouter {...routerParams}>
        <Site />
      </StaticRouter>
    ))
    .then(payload => {
      const {
        data: resolvedData,
        Resolved: ResolvedApp,
      } = payload;

      const sheets = new SheetsRegistry();

      const content = ReactDOMServer.renderToString(
        <JssProvider registry={sheets}>
          <ResolvedApp />
        </JssProvider>
      );

      const helmet = Helmet.renderStatic();

      ctx.body = html`
        <!doctype html>
        <html ${helmet.htmlAttributes.toString()}>
          <head>
            ${helmet.meta.toString()}
            ${helmet.title.toString()}
            ${helmet.link.toString()}
            ${helmet.style.toString()}
            <style>${sheets.toString()}</style>
          </head>
          <body ${helmet.bodyAttributes.toString()}>
            <div data-placeholder="site-content">${content}</div>
            ${bootstrapData(DATA_RESOLVER_BOOTSTRAP_KEY, resolvedData)}
            ${helmet.script.toString()}
            ${helmet.noscript.toString()}
          </body>
        </html>
      `;
}
```

As you can see I have two renders:

1. For async data resolution
1. For final render with resolved data

a-ignatov-parc:
If I'll wrap first render with `JssProvider` then I'll get duplicate styles

kof:
The use case is basically 2 render calls for one document. If you use 2 JssProvider you get conflicting class names, because the counter will be reset and you can't use one JssProvider for both because they happen completely separate.

kof:
I think in this case, your tests don't exactly cover your use case.

kof:
For e.g. class name could contain a provider id so that each provider ensures uniqueness of classes for its children.

a-ignatov-parc:
Once component rendered without `JssProvider` providing custom registry on next render won't reset counters.

a-ignatov-parc:
And as I mentioned in description everything was ok in `react-jss@7.0.1`.

kof:
> Once component rendered without JssProvider providing custom registry on next render won't reset counters.

exactly, thats why you always need to use JssProvider on the server.

kof:
> And as I mentioned in description everything was ok in react-jss@7.0.1.

Wait, are you saying it broke in 7.0.2 but was fine in 7.0.1?

a-ignatov-parc:
> Wait, are you saying it broke in 7.0.2 but was fine in 7.0.1?

Exactly

kof:
Not sure how, but we actually fixed a very embarrassing bug in 7.0.2. Maybe it worked because it was so terribly broken :)

a-ignatov-parc:
That isn't very good news ))

kof:
Do we agree that JssProvider should be always used on the server? If so the test case would be to use it twice and not have conflicting class names.

kof:
It might bring a different problem though if on the client provider isn't used exactly the same way.

a-ignatov-parc:
I see [cached `SheetsManager` instance](https://github.com/cssinjs/react-jss/blob/master/src/createHoc.js#L52). It will be shared beetween renders and could be the cause for my problem.

kof:
It is in the scope in order to reuse the same static sheet between all elements.

a-ignatov-parc:
> It is in the scope in order to reuse the same static sheet between all elements.

So this cached instance will share styles between renders because it's not aware about active registry?

kof:
> So this cached instance will share styles between renders because it's not aware about active registry?

It will if you don't use JssProvider.

a-ignatov-parc:
Here is a fix for shared static sheet. Now it has bound to registry.

src/createHoc.js
```diff
@@ -48,9 +48,8 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
   const isThemingEnabled = typeof stylesOrCreator === 'function'
 
   const displayName = `Jss(${getDisplayName(InnerComponent)})`
+  const staticSheetCache = new Map()
```
a-ignatov-parc: I haven't found any info about supported browsers by `react-jss`. We need to be careful with `Map`, it has support starting from IE 11

kof: same as react, IE9+

kof: Thats why I simulate Map using an array and indexOf

a-ignatov-parc: Ok, get it. Will change to custom implementation

a-ignatov-parc: Done

src/map.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Minimal implementation of the Map data structure for old browsers.
+ * https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map
+ */
+class CustomMap {
+  constructor() {
+    this.index = []
+    this.content = {}
```
iamstarkov: its should be an array

iamstarkov: and do we need tests for this?

src/index.test.js
```diff
@@ -364,6 +364,48 @@ describe('react-jss', () => {
...
+        <JssProvider jss={localJss} registry={customSheets2}>
+          <Component color="#000" />
+        </JssProvider>
+      )
+
+      // Removing appended styles after `ReactDOMServer.renderToString` because
+      // we can't use virtual renderer for this specific test case.
+      Array
+        .from(document.querySelectorAll('style'))
+        .forEach(styleNode => styleNode.parentNode.removeChild(styleNode))
```
iamstarkov: are those `styleNode`s have been attached? `renderToString` shouldnt do that. Does it do that? /cc @kof

iamstarkov: how does this line work? https://github.com/cssinjs/jss/blob/master/src/StyleSheet.js#L57

src/index.test.js
```diff
@@ -364,6 +364,48 @@ describe('react-jss', () => {
...
+
+      renderToString(
+        <Component color="#000" />
+      )
+
+      renderToString(
+        <JssProvider jss={localJss} registry={customSheets2}>
+          <Component color="#000" />
+        </JssProvider>
+      )
```
iamstarkov: why do you render it twice, if your failed use case requires just two renders?

src/map.js
iamstarkov: i think we definitely need tests for this

iamstarkov:
reviewed: request for more tests and some questions about renders

src/createHoc.js
kof: I think we can have a way more minimalistic replacement for map if we don't try to abstract it nicely.

src/map.js
```diff
@@ -0,0 +1,36 @@
+/**
```
kof: As far as I can see, we can do what we need in 2 LOC instead of 36

src/createHoc.js
iamstarkov: thanks for the fix.

im struggling to justify map of a map (`staticSheet` map by `registry` id of `Sheets` map by `theme` key)

btw, HMR will reset every `createHoc`'s variable and this particular `staticSheetCahe` map

src/map.js
a-ignatov-parc: Object is more simpler data structure than array also it more performant on remove operations (`O(1)` instead `O(n)` for array).

src/index.test.js
a-ignatov-parc: Need to check this. It was fast attempt to create failing test case

src/index.test.js
a-ignatov-parc: `renderToString` without virtual rendering is attaching styles in browser

src/map.js
kof: Sure, I just don't think this kind of optimization is needed here.

src/index.test.js
iamstarkov: @kof, how do we pass virtual renderer to jss?

src/map.js
a-ignatov-parc: By not using Map at all? 

I don't think that it would be a good idea because of Map data structure and its algorithmic complexity for get/add/remove operations (only `O(1)`).

src/index.test.js
a-ignatov-parc: > how do we pass virtual renderer to jss?

By [passing `virtual` param](https://github.com/a-ignatov-parc/react-jss/blob/503b79fcf9ee62b74f1f6f395f851986f04d797a/src/index.test.js#L232)

src/map.js
a-ignatov-parc: And because there is no difference between object and array in this specific case let it be object if you don't mind ;)

kof: I can do the same with just one array in 2 LOC instead of 26. So either we do this or we go for an external Map polyfill requirement. I prefer the first one.

kof:
We need to think what to do with HMR. This one will break the fix in the other PR as far as I know.

src/map.js
a-ignatov-parc: @kof maybe you can share your code? I don't see how I can reduce size of this module to 2 LOC and remain with compatible `Map` api 🤔

a-ignatov-parc:
> We need to think what to do with HMR. This one will break the fix in the other PR as far as I know.

Are you talking about this PR https://github.com/cssinjs/react-jss/pull/123?

src/index.test.js
a-ignatov-parc: Removed extra render

a-ignatov-parc:
Checked this fix with the one in https://github.com/cssinjs/react-jss/pull/123 
It works fine as I can say :)

a-ignatov-parc:
Added tests + made small code refactoring

src/objectMap.js
```diff
@@ -0,0 +1,50 @@
+/**
+ * Key/value storage with ability to store objects as keys.
+ * Uses `Map` for performance reasons with fallback for old browsers.
+ */
+export default function objectMap(params) {
+  const indices = []
+  const content = []
+  const map = !(params && params.fallback)
+    && typeof Map !== 'undefined'
+    && new Map()
```
a-ignatov-parc: I think it would be good idea to change `Map` to `WeakMap` to help GC during SSR 🤔

src/index.test.js
iamstarkov: so, perhaps we should use it

a-ignatov-parc: We can't provide any params to this particular instance because the use case is to not to use `JssProvider` :)

But maybe we can set `virtual` param globally?

a-ignatov-parc:
Any news on this PR?

iamstarkov:
@kof lets not wait until #123

iamstarkov:
its separate issue, and shouldnt block this one

kof:
This PR gives me headache. Lets have a chat on gitter.

kof:
We discussed the issue and decided to go for a different solution, see https://github.com/cssinjs/react-jss/issues/133

Issue #100, opened 2017-06-20
Add component name to the generated class name 


kof:
To not force people switch between react dev tools and dom inspector. Reported by @rob2d

kof:
This should wait for JSS 8 comming out soon, because theere will be a breaking change of generateClassName option.

iamstarkov:
are we ready?

kof:
yep, lets do this)

rob2d:
Awesome. Since I'm the one who brought it up/requested this, would you need any help on it?

iamstarkov:
@rob2d if you can add tests, that would be nice

iamstarkov:
@rob2d what do you think about it?

rob2d:
Sorry about the delay. I can try to get that in early this weekend. 🙂

iamstarkov:
thanks

iamstarkov:
that would be cool

rob2d:
Added a test for generated classnames to contain the Component's `name` property 😃 

Please let me know if there's any issues with that.

kof:
next jss version will support classNamePrefix option

rob2d:
awesome! 😁 looking forward to it. thanks for adding this feature.

kof:
Released v8

PR #125, opened 2017-07-23
Add test for component name in classname

rob2d:
references issue #100. Just adding a test which reflects the planned functionality to add a component's calling name before it's className --e.g. `MyComponent-container-0-1`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-jss",
-  "version": "7.0.1",
+  "version": "7.0.2",
   "description": "Integration of JSS for react.",
   "main": "lib/index.js",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-jss",
-  "version": "7.0.2",
+  "version": "7.1.0",
   "description": "Integration of JSS for react.",
   "main": "lib/index.js",
   "engines": {
--- a/src/index.test.js
+++ b/src/index.test.js
@@ -125,9 +125,9 @@ describe('react-jss', () => {
 
     it('should be prefixed by the parent injected component\'s name', () => {
       render(<Component />, node)
-      Object.keys(passedClasses).forEach((className) => {
-        classNameRegex = new RegExp(`^${InnerComponent.name}-${className}((-)?[\\d]*)*$`)
-        expect(passedClasses[className]).to.match(classNameRegex)
+      Object.keys(passedClasses).forEach((ruleName) => {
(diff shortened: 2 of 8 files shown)


src/index.test.js
```diff
@@ -120,6 +122,14 @@ describe('react-jss', () => {
       render(<Component classes={classes} />, node)
       expect(passedClasses).to.equal(classes)
     })
+
+    it('should be prefixed by the parent injected component\'s name', () => {
+      render(<Component />, node)
+      Object.keys(passedClasses).forEach((className) => {
+        classNameRegex = new RegExp(`^${InnerComponent.name}-${className}((-)?[\\d]*)*$`)
```
kof: - Feel free to use .displayName or even hardcode the name.
- Your tests says its all about the name prefix, but the regex seems to verify more. Also what you call className is a rule name, class name is a value in that map.

kof: instead of generating a regex it would be easier to run an expect call in that forEach  or even simpler - just use that one property, it is ok to be hardcoded, its a test.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-jss",
-  "version": "7.0.1",
+  "version": "7.0.2",
```
kof: pull requests should not have any opinions about the version number as we don't know when it will be released at merge time.

src/index.test.js
rob2d: The RegExp isn't too much work for me but good point :) I can loosen up the check for just the prefixing. I've also noticed that many people tend to forget to add `displayName` to their component, but ES6's spec says function vars have the `name` property by default prepended. I was trying to model that assumption. It makes complete sense to prioritize `displayName` over `name` though and can omit completely if you insist.

rob2d: will make these adjustments after work today or during a break. thanks!

package.json
rob2d: sorry, this was actually just a stupid mistake pulling from another commit which was made on the repo (just installed GitKraken whereas I usually use SourceTree). I will try to be more careful with that going forward.

rob2d:
alrighto -- that was a bit later than expected but I just made adjustments to the test that was written. Now prioritizes displayName and does not restrict the regex input at all after `ComponentName-ruleName`

package.json
kof: any update?

kof:
merged, thanks!

rob2d:
No problem. Thank _you_ for that. Been looking forward to this feature 😄 👍

Issue #135, opened 2017-08-02
Allow themeListener with custom channel


catalin-enache:
Hello react-jss team,

I have the following Scenario:

Lib A uses ```react-jss``` for its UI components.
Lib B uses ```react-jss``` for its UI components.
Dev X wants to use a component from lib A and a component from lib B using different ```ThemeProviders```.

As react-jss currently uses ```themeListener``` straight from ```theming``` (which channel defaults to ```__THEMING__ ```) is not possible
for lib A and lib B to choose their custom channel.

If this would be possible we could have something like:

```
<ThemeProviderA>
  <ThemeProviderB>
    <LibAComponent/>
    <LibBComponent/>
  </ThemeProviderB>
</ThemeProviderA>
```

What do you think of allowing ```injectSheet``` to have as a third param the ```themeListener``` which should default to ```themeListener``` from ```theming``` ? Then ```injectSheet``` passes down ```themeListener``` to ```createHOC```.

This way lib A and lib B could setup their ```ThemeProviders``` and ```themeListeners``` with custom channels by using ```createTheming('__LIB_A_THEMING__')```, ```createTheming('__LIB_B_THEMING__')``` from ```theming``` package, then they can pass inside ```injectSheet``` the custom ```themeListener``` resulted from calling ```createTheming``` function.

Thank you.

kof:
Yeah, sounds like a reasonable request.

catalin-enache:
Hello @kof, thank you for feedback.
I created a branch on my local machine for this feature ( + unit-test)
The changes are as suggested in my comment.
I'd like to push it for review somehow.
How can I do it without pasting the diff here
which would make this thread super long and unreadable ?
Thank you.

kof:
How about a pr?

catalin-enache:
Hi @kof,
I forked react-jss and created a PR from the fork:
https://github.com/cssinjs/react-jss/pull/136

For whatever reasons the build failed:
"No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)"

However, before pushing the PR I ran "lint" and "test" and they were successful.

iamstarkov:
@catalin-enache we dont have configured travis right now, so no worries

kof:
this has been published

catalin-enache:
Already using it. Thank you.

PR #136, opened 2017-08-04
allow theming sandbox via custom themeListener

catalin-enache:
allow theming sandbox via custom themeListener. if PR is OK I can update the docs too.

### DIFF
--- a/src/createHoc.js
+++ b/src/createHoc.js
@@ -1,5 +1,5 @@
 import React, {Component} from 'react'
-import {themeListener as _themeListener} from 'theming'
+import {themeListener as defaultThemeListener} from 'theming'
 import jss, {getDynamicStyles, SheetsManager} from './jss'
 import compose from './compose'
 import getDisplayName from './getDisplayName'
@@ -42,17 +42,14 @@ const getStyles = (stylesOrCreator, theme) => {
  * @param {Object|Function} stylesOrCreator
  * @param {Component} InnerComponent
  * @param {Object} [options]
- * @param {Object} [themeListener]
  * @return {Component}
  */
-export default (
-    stylesOrCreator,
-    InnerComponent,
-    options = {},
-    themeListener = _themeListener
-) => {
+export default (stylesOrCreator, InnerComponent, options = {}) => {
   const isThemingEnabled = typeof stylesOrCreator === 'function'
-
+  const {
+    themeListener = defaultThemeListener,
+    ...sheetOptions
+  } = options
   const displayName = `Jss(${getDisplayName(InnerComponent)})`
   const noTheme = {}
   let manager = new SheetsManager()
@@ -94,7 +91,7 @@ export default (
       if (!staticSheet) {
         const styles = getStyles(stylesOrCreator, theme)
         staticSheet = this.jss.createStyleSheet(styles, {
-          ...options,
+          ...sheetOptions,
           ...this.context[ns.sheetOptions],
           meta: `${displayName}, ${isThemingEnabled ? 'Themed' : 'Unthemed'}, Static`
         })
@@ -106,7 +103,7 @@ export default (
 
       if (dynamicStyles) {
         dynamicSheet = this.jss.createStyleSheet(dynamicStyles, {
-          ...options,
+          ...sheetOptions,
           ...this.context[ns.sheetOptions],
           meta: `${displayName}, ${isThemingEnabled ? 'Themed' : 'Unthemed'}, Dynamic`,
           link: true
--- a/src/createHoc.js
+++ b/src/createHoc.js
@@ -47,9 +47,10 @@ const getStyles = (stylesOrCreator, theme) => {
 export default (stylesOrCreator, InnerComponent, options = {}) => {
   const isThemingEnabled = typeof stylesOrCreator === 'function'
   const {
-    themeListener = defaultThemeListener,
+    theming = {},
     ...sheetOptions
   } = options
+  const {themeListener = defaultThemeListener} = theming
   const displayName = `Jss(${getDisplayName(InnerComponent)})`
   const noTheme = {}
   let manager = new SheetsManager()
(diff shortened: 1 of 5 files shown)


catalin-enache:
This PR is related to: https://github.com/cssinjs/react-jss/issues/135

src/createHoc.js
```diff
@@ -42,9 +42,15 @@ const getStyles = (stylesOrCreator, theme) => {
...
  * @param {Object} [options]
+ * @param {Object} [themeListener]
  * @return {Component}
  */
-export default (stylesOrCreator, InnerComponent, options = {}) => {
+export default (
+    stylesOrCreator,
+    InnerComponent,
+    options = {},
+    themeListener = _themeListener
```
kof: 4 arguments is def. a bad design

kof: Lets put themeListener into options. And lets not use "_" to avoid collisions.

kof: Another option is to just add a "channel" option and let jss make createTheming call. This way user doesn't need to add theming a dependency. There is no any other options in createTheming, so we loose nothing and avoid more problems.

kof: Also "channel" name used in the theming package is probably not very accurate. That thing is just a namespace, not a "channel". Channel would be the emitter. cc @iamstarkov

kof: In react-jss options context I would call it "themingContextNamespace"

kof: maybe we could shorten it to "themingNamespace"

catalin-enache: I was also thinking of _"just add a "channel/namespace" option and let jss make createThemeListener "_. As you suggested, this seems indeed more elegant . But the user still need to match the ```themeListener``` with corresponding ```ThemeProvider``` (on the same channel/namespace). He needs to get the ```ThemeProvider``` for his custom channel/namespace from somewhere (either from ```theming``` directly or from ```react-jss``` indirectly). Would you like this approach ? Let ```react-jss``` EXPOSE/FORWARD ```createTheming``` from ```theming``` library. User makes this call: ```const {ThemeProvider, [optional themeListener] } = createTheming('__MY_NAMESPACE__');``` (where ```createTheming``` was exposed by ```react-jss``` this time). User uses from now on this ```ThemeProvider``` that ```createTheming``` returned to wrap his application. User later calls ```injectSheet``` providing as an option either the ```themeListener``` object (which was also returned from ```createTheming```) or just the ```channel/namespace string```. ```react-jss``` will need to maintain internally a registry mapping channels/namespaces to ThemeProviders/themeListeners and stuff. What do you think ? Thank you.

kof: yeah, forgot about ThemeProvider ... it sounds like a lot of complexity we currently have in order to create a custom namespace.

kof: We can expose `createTheming` function from react-jss. It creates both ThemeProvider and  themeListener. We can accept that object `theming` in createHoc. 

Another q is how would you use injectSheet? You would need to pass that theming object to every injectSheet call or wrap it.

kof: maybe we need something like `createReactJss` which accepts theming among other things and returns ThemeProvider, JssProvder, injectSheet etc

catalin-enache: I would keep it simpe and go with your previous suggestion (not this/latest one).  expose ```createTheme```  and accept in ```injectSheet``` the ```themeListener``` resulted from calling ```createTheme```. The user can create a wrapper over ```injectSheet``` that should take care to always inject ```themeListener``` and forget about ```injectSheet``` after that. This approach avoids complicating react-jss with the the burden of taking care of custom theme. We can write an example of how to do it and let the user taking care of details. What do you think ?

catalin-enache: More than that, theming is a third party feature. Why handling this internally?

kof: I was thinking about other use cases like that and thinking about a more generic solution.

kof: But yeah, for now probably [this](https://github.com/cssinjs/react-jss/pull/136#discussion_r131544598) would be enough.

catalin-enache: Please check the last commit with changes according to previous suggestions.
Regarding one of your comments _"And lets not use "_" to avoid collisions"_, the linting wont allow that.
I just used another alias _"defaultThemeListener"_
If is OK I'll update the docs too. Thank you.

kof: maybe accept `theming` object you received from createTheming? so that user knows less about internals like themeListener.

src/createHoc.js
```diff
@@ -46,7 +46,11 @@ const getStyles = (stylesOrCreator, theme) => {
  */
 export default (stylesOrCreator, InnerComponent, options = {}) => {
   const isThemingEnabled = typeof stylesOrCreator === 'function'
-
+  const {
+    theming = {},
```
kof: you could default to theming object from theming package and import on line 2 theming, right?

kof:
ok, lets add the docs and we are good.

src/createHoc.js
catalin-enache: Not sure I got this. It's another way of saying:
```
const themeListener = (options.theming || {}).themeListener || defaultThemeListener
```

kof: import theming from 'theming'
const themeListener = (options.theming || theming).themeListener

kof:
merged, thanks.

catalin-enache:
I thank you, also for post editing. There is a small issue now in the docs. ```theming``` object passed inside ```injectSheet``` options does not exist in that context. You deleted the line which kept the ```theming``` object in its variable to be passed/used later, by shrinking 2 lines into one line.

kof:
ups, thanks

catalin-enache:
When is it planned to push a new release to npm registry ? Using git link "react-jss": "cssinjs/react-jss" in package.json dependencies does not work currently as "src" directory is not retrieved (I guess because is not listed in package.json "files" entry).

kof:
was hoping to get more stuff done in this release, but released right now just this in 7.1.0

catalin-enache:
Very much appreciated. Thank you.

Issue #133, opened 2017-07-28
JssProvider enhancements


kof:
-  [x] 1. Do NOT save manager in the scope. Async requests will use it because injectSheet() is called just once.
- [x] 2. Warn when JssProvider is used inside of JssProvider (use context). Update: no need, see https://github.com/cssinjs/react-jss/issues/157#issuecomment-333576671
- [x] 3. No JssProvider - no styles. For this we can pass jss from Provider, no jss no styles. This can be used legally in order to render components without styles for e.g. in order to resolve some async  deps on the server. Warn on the client if no JssProvider is used.
- [x] 4. Use registry object instead of random providerId to identify the same request
- [x] 5. Add a bold hint in readme to NOT use 2 JssProviders in parallel for one request in separate contexts, because it will result in conflicting class names???? We need to research if there are use cases for a strong need of multiple parallel JssProviders. If there are, we need to think of something to support it. Update: we just found a [valid use case](https://github.com/callemall/material-ui/issues/8223#issuecomment-331081785)

    - [x] add generateClassName prop 
    - [x] add classNamePrefix prop

oliviertassinari:
2. 💯 
5. I have noticed a similar issue on the Material-UI side, where people have conflicting class names. I haven't had the opportunity to dig into it. My point is just that counter base solution is more challenging to get right 😬 .

tiagoroldao:
Regarding 5 - I'm running into an issue where multiple contexts are unavoidable and break stuff: Atom view development (in particular for use in panels - i.e. specific file type editing)

When opening/closing everything is fine, but when multiple tabs are open they are essentially separate apps with separate contexts within them, and the manager breaks.

tiagoroldao:
A couple of questions to see if I understand this project better and can propose solutions:
- why does JssProvider provide a providerId to the context, but not a manager?
- using providerId, could the "static" var `manager` in `createHoc` become a map of providerId's to managers?

kof:
For SSR, we need to identify a new http request. Provider is rendered only once at each request, thats how we identify requests.

Manager is responsible for handling sheets of one component. It makes sure a static sheet is reused across all component instances. Also it is used to ensure that components which use the same theme use the same sheets, but not different themes.

tiagoroldao:
Two possible solutions to 5:

tiagoroldao:
Two possible solutions:
[...]

PR #148, opened 2017-09-06
Define manager in context through Provider when used

tiagoroldao:
This to allow multiple parallel contexts to coexist. Alternate to #149


tiagoroldao:
Relates to #133

tiagoroldao:
Manual testing in use in Atom Editor (the mentioned use case for multiple contexts) shows this **not to work**. #149 appears to work properly, defining a manager per Component, per context.

Issue #132, opened 2017-07-25
injectStylesheet: merge user classes property


oliviertassinari:
This issue is part of a series of feedbacks. [I been upgrading the Material-UI solution](https://github.com/callemall/material-ui/pull/7461).
Unfortunately, I had to fork the package in order to reach my goal.
Now, it's time to reconciliation.

### The issue

Regarding that issue, I don't know if that fit with the vision of the project. I'm opening it in the doubt.
So, I'm suggesting to implement the `classes` property logic of Material-UI in react-jss.

kof:
@oliviertassinari can you please elaborate with more detail?

oliviertassinari:
Here is the [Material-UI case](https://material-ui-1dab0.firebaseapp.com/customization/overrides#overriding-with-classes) for the classes property.
That's always something that Material-UI could build on top of createHOC.js

kof:
Can't see whats special in mui in this regard. Does it simply allow to override classes object by user props? If so we already support this.

oliviertassinari:
@kof, it's not a simple override. Instead, we merge the keys.

kof:
Q: the merge you do in mui, would it override user component class names if they match? or it acts like a `defaults` function?

oliviertassinari:
We merge them. We don't override it.

kof:
Yes, but what direction does merge goes: merge(HocClasses, ChildClasses) or the other way around?

oliviertassinari:
Does it matter? We concatenate the classe names.

kof:
Oh so you merge not just the objects, you merge the class names strings!

oliviertassinari:
Yes indeed :).

kof:
1. Create a classes object and put it into instance
2. Merge in classes strings from the dynamicSheet | sheet
3. Merge in classes strings from props.classes

jedwards1211:
@kof how would this behave if `defaultProps.classes` are copied from the wrapped component though?  Wouldn't the classes strings from `defaultProps.classes` still end up overwriting the strings from the `dynamicSheet | sheet` (assuming I don't specifically pass `classes={{...}}` when I create an instance of the wrapper?)

kof:
Actually I was going to merge classes always. So it would be always consistently merged.

What bad can happen if `props.classes` will be merged with your default classes where you wanted them to be used only if JSS HOC was not used?

```javascript
classes = {
  button: 'default-class-name generated-class-name'
}
```

[...]

Issue #87, opened 2017-04-13
Default props are not passed to sheet.update()


kserjey:
Default props is not injected in update data.
[Example](https://www.webpackbin.com/bins/-KhaeerXIdMPyBidaJ9r)

kof:
sounds like  a bug

kof:
Its clear now - default props are defined on the inner component. Not on the hoc.

kof:
cc @iamstarkov wdyt? should we merge defaultProps from the user component to the HOC? We currently use "hoist-non-react-statics" package which does this for all other props.

kof:
fixed and published v6.1.0

jedwards1211:
@kof this exact same issue was filed in `react-redux` and they **don't** consider it a bug: https://github.com/reactjs/react-redux/issues/111

As they said in that issue, if @kserjey wants a default `backgroundColor` to get passed to functions in `baseStyles`, he should explicitly add `defaultProps` to the component created by `injectSheet(baseStyles)(SomeComponent)`, not to `SomeComponent` itself.

kof:
Yep, got it now, will try to fix both issues.

jedwards1211:
Cool, thanks!

PR #153, opened 2017-09-18
fix(createHoc): don't copy defaultProps from wrapped component

### LINKED ISSUE #152: If innerComponent has `defaultProps.classes`, they override `classes` injected by HOC
I don't really understand why the HOC copies over so much stuff like `defaultProps` and `contextTypes` from the wrapped component instead of just creating a standalone component that renders the wrapped one.  If I have

```js
class MyComp extends React.Component {
  static defaultProps = {
    classes: {
      root: 'MyComp-root',
    }
  }
}

const styles = {
  root: {},
}

export default injectSheet(styles)(MyComp)
```

The created HOC doesn't work because it takes on the `defaultProps` from `MyComp`, so the rendered `classes.root` is always `'MyComp-root'`. 💥 💣

jedwards1211:
fix #152

### DIFF
--- a/src/createHoc.js
+++ b/src/createHoc.js
@@ -60,7 +60,6 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       ...contextTypes,
       ...(isThemingEnabled && themeListener.contextTypes)
     }
-    static defaultProps = InnerComponent.defaultProps
 
     constructor(props, context) {
       super(props, context)


kof:
That line was there for  a reason)

kof:
Thanks for the PR, closing this in favor of https://github.com/cssinjs/react-jss/issues/132

jedwards1211:
What was the reason?  It's the kind of thing where my initial prejudice is whatever that reason might be, I believe there has to be a better way to do it

kof:
Do you have another use case in mind that demonstrates its not a good idea to forward default props to the HOC? Imagine classes merge is already fixed.

jedwards1211:
Yes, creating a library component that users can use with plain CSS or with JSS:

`CircularProgress.js`
```js
export default class CircularProgress extends React.Component {
  static defaultProps = {
    // class names for when the user is importing a CSS file -
    // if the user wraps this with `injectSheet`, these should not get used.
    classes: {
      root: 'CircularProgress-root',
      track: 'CircularProgress-track',
      ...
    }
  }
}
export const jssStyles = {
  root: {
    ...
  },
  track: {
    backgroundColor: '#eeeeee',
  },
}
```

`CircularProgress.css`
```css
.CircularProgress-root {
  ...
}
.CircularProgress-track {
  background-color: #eeeeee;
}
```

`CircularProgress-jss-styles.js`
```js

```

Usage with CSS
```js
import 'react-circular-progressbar/CircularProgress.css'
import CircularProgress from 'react-circular-progressbar'
```

Usage with JSS
```js
import BaseCircularProgress, {jssStyles} from 'react-circular-progressbar'
const CircularProgress = injectSheet(jssStyles)(BaseCircularProgress)
```

jedwards1211:
Hang on, the example I posted was not very good...

jedwards1211:
Okay, imagine you want to create a `Panel` component that uses Bootstrap's panel classes by default.  But if you want to style it differently, you can pass your own `classes` prop to it and it will use those instead of Bootstrap CSS.

```js
const Panel = ({classes, header, footer, children}) => (
  <div className={classes.panel}>
    {header && <div className={classes.header}>{header}</div>}
    {children && <div className={classes.body}>{children}</div>}
    {footer && <div className={classes.footer}>{footer}</div>}
  </div>
)
Panel.defaultProps = {
  classes: {
    panel: 'panel',
    header: 'panel-header',
    body: 'panel-body',
    footer: 'panel-footer',
  }
}

<Panel header={<h4>Bootstrap Panel</h4>}>Body</Panel>

const styles = {
  panel: {...},
  header: {...},
  body: {...},
  footer: {...},
}
const CustomPanel = injectSheet(styles)(Panel)

<CustomPanel header={<h4>Custom Panel</h4>}>Body</CustomPanel>
```

So `<Panel>` above should render
```jsx
<div className="panel">
  <div className="panel-header">
    <h4>Bootstrap Panel</h4>
  </div>
  <div className="panel-body">
     Body
  </div>
</div>
```

And `<CustomPanel>` above should render something like (I forget how the numbering works):
```jsx
<div className="panel-0-0">
  <div className="header-0-1">
    <h4>Bootstrap Panel</h4>
  </div>
  <div className="body-0-2">
     Body
  </div>
</div>
```

But with `defaultProps.classes` getting merged in, we'd get Bootstrap styles clashing with the injected ones:
```jsx
<div className="panel panel-0-0">
  <div className="panel-header header-0-1">
    <h4>Bootstrap Panel</h4>
  </div>
  <div className="panel-body body-0-2">
     Body
  </div>
</div>
```

kof:
Is that scenario realistic?

kof:
@oliviertassinari wdyt?

jedwards1211:
@kof somewhat.  Here's another question though: why should rendering an instance of an `injectSheet(styles)(Panel)` behave any differently than rendering `<Panel classes={{panel: 'panel-0-0', header: 'header-0-1', ...}} />`

oliviertassinari:
I don't follow, this PR looks great.

oliviertassinari:
Also, I would add [hoist-non-react-statics](https://github.com/mridgway/hoist-non-react-statics)

kof:
@oliviertassinari this is already in place.

jedwards1211:
In any case, if `defaultProps` aren't automatically copied, it gives people more flexibility, since they still have the option of copying over `defaultProps` manually if they want

kof:
PR is not good, because there is a failing test. I am looking into it.

kof:
Now as I looked into the failing test, the reason was a different one. It is the passing props object to a function value. It was basically an easy way to do it, because I just reference it and react does the merge, but now I see that it is actually wrong and needs fixing.

jedwards1211:
@kof this does re-break what @kserjey expected in https://github.com/cssinjs/react-jss/issues/87
But, over there I argue that `injectSheet` should be working like `react-redux` and #87 shouldn't be considered a bug.

Issue #157, opened 2017-10-01
Allow nesting of JssProvider


oliviertassinari:
The current implementation of the JssProvider makes it impossible to gradually provide configuration options in the render tree.
https://github.com/cssinjs/react-jss/blob/cda37731dad5e1ecdea61ec6eabb62500a169383/src/JssProvider.js#L32-L39

I have a use case with next.js where
- I'm using the `sheetsRegistry` option to collect the server side generated side:
https://github.com/callemall/material-ui/blob/v1-beta/pages/_document.js#L104-L106
- On the client side, I want to provide a `jss` instance to customize the injection point https://github.com/callemall/material-ui/blob/v1-beta/docs/src/modules/components/AppWrapper.js#L13-L22

oliviertassinari:
Actually, I think that I have a workaround where a single `JssProvider` is enough. I will post the commit here once it's confirmed.

oliviertassinari:
Here is my workaround: https://github.com/callemall/material-ui/pull/8497. Not bad. As far I'm concerned it's no longer an issue, but extra flexibility won't harm. Your call.

oliviertassinari:
Oh no, damn, it's not working 🔧

oliviertassinari:
Ok, I have been making a stupid mistake, we are back on track.

kof:
You could use one JssProvider on the server and a different one on the client, right?

kof:
Supporting nesting providers will lead to a lot of complexities and indirection.

kof:
Ok I got a plan:
1. On the client, JssProvider should not  reset the managers when rerendered
2. It should forward context props defined by the parent JssProvider if no specific new prop has been defined. In this case it works like a proxy.

oliviertassinari:
I like the option of being able to incrementally add key into the context. Regarding overriding one that is already present. You could warn if it's simpler or extend. Your plan seem to go into this direction 👍 .

kof:
I think if user wants to override some setting for e.g. use a different Jss instance for a subtree, we should let him do  that.

kof:
Released v8

PR #160, opened 2017-10-02
Fixes manager for SSR

kof:
Fixes (1) in #133


kof:
cc @oliviertassinari

oliviertassinari:
@kof I'm not sure my feedback is relevant here. I have been adopting a different solution to handle style deduplication on Material-UI: the `sheetManager` is a simple `Map`.

kof:
I am assuming you can switch step by step to use react-jss as is in the future. So it is important that I don't do anything that mui can't use.

kof:
We have a more complex scenario because of theming and dynamic styles.

oliviertassinari:
@kof Oh, then we need to discuss this sheetManager further on gitter. This is the part I had to do the most changes. I don't think that the current implementation will satisfies Material-UI needs.

kof:
Alright, now #157 is also implemented.

kof:
Ok, this one seems ready

PR #161, opened 2017-10-04
Added inject option

kof:


kof:
cc @iamstarkov @oliviertassinari

src/createHoc.js
```diff
@@ -36,6 +36,18 @@ const getStyles = (stylesOrCreator, theme) => {
   return stylesOrCreator(theme)
 }
 
+// Returns an object with array property as a key and true as a value.
+const toMap = arr => arr.reduce((map, prop) => {
+  map[prop] = true
+  return map
+}, {})
+
+const defaultInjectMap = {
```
oliviertassinari: Have you considered the following values? It's what I'm using in 80%+ of the time. It would be nonoptimal if I had  to destructure `sheet` and `theme` every time I want to use the HOC. I most of the time spread the props on the root component (in userland, I'm not talking about Material-UI, but it applies to Material-UI too).
```jsx
{
  sheet: false,
  classes: true,
  theme: false
}
```

src/createHoc.js
```diff
@@ -189,9 +202,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
     render() {
       const {theme, dynamicSheet} = this.state
       const sheet = dynamicSheet || this.manager.get(theme)
-      const reactJssProps = {sheet, classes: sheet.classes}
-      if (isThemingEnabled) reactJssProps.theme = theme
-      return <InnerComponent {...reactJssProps} {...this.props} />
+      const jssProps = {}
+      if (injectMap.sheet) jssProps.sheet = sheet
+      if (injectMap.classes) jssProps.classes = sheet.classes
+      if (isThemingEnabled && injectMap.theme) jssProps.theme = theme
```
oliviertassinari: It's funny how I have an impression of déjà vu. So, this was my first implementation on Material-UI side, until I realized that I had a component no using a function style but requiring the theme. I think that you should update your `isThemingEnabled` implementation.

kof: hmm, not sure I understand what you are saying)

src/createHoc.js
kof: I agree regarding sheet, I have no use for it inside of the components. On the other hand if theme is only passed when ThemeProvider is used. But yeah, most of the time, theme should be needed for the styles.

I am wondering if we need a global way to change that.

src/createHoc.js
oliviertassinari: I'm suggesting the following:
```
 const isThemingEnabled = typeof stylesOrCreator === 'function' || injectMap.theme
```

kof: In this case injectMap.theme is not just a filter, it is a switch.

kof: Also I don't know if there is much use for theme in the render fn when styles is not using a theme.

oliviertassinari: I used to have one on Material-UI, it's no longer true. In that case, you can raise a warning. Suggesting to use `withTheme` instead.

kof: Its possible, adds a lot of  complexity though.

src/createHoc.js
kof: For MUI this is actually not a big deal as it can easily wrap and setup once for all. For regular users it depends on whether a user uses with or without a theme most of the time. I guess we don't know that. The only thing that I am 90% sure is that most cases don't need sheet prop.

kof:
Yeah, we need to think though how to do without introducing maintenance burden.

kof:
removed sheet prop by default

PR #187, opened 2017-12-18
Unsubscribe when the component un mounts

TriPSs:
I received a lot of `setState` warnings in my console and noticed that the components did not remove their listeners when un mounting.

This PR goes together with [theming - 48](https://github.com/cssinjs/theming/pull/48)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4384,6 +4384,15 @@
             }
           }
         },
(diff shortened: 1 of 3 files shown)


kof:
thanks!

kof:
published

PR #188, opened 2017-12-19
Update README.md

davesidious:
Touched up the copy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# React integration of JSS
+# React integration with JSS
 
 [![Gitter](https://badges.gitter.im/JoinChat.svg)](https://gitter.im/cssinjs/lobby)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# React integration of JSS
+# React integration with JSS
```
kof: not sure why "with", react-jss is literally integration of JSS for react

davesidious: It's because one usually integrates *with* something. Implementation would be the word to use if one wanted "of", I'd imagine.

kof: Then it should be JSS integration with React, no?

davesidious: You raise a good point!  As integration implies a mutual endeavour, so I'm not entirely sure the ordering is gramatically important, but as the subject of this document is more JSS-specific, it would make sense to have that first.  Should I make this change?

kof: yes, please, then I can just merge it)

kof: same might be in package.json

kof:
thanks a lot!

PR #190, opened 2017-12-28
document how to inject custom generateClassName

jedwards1211:
not sure if you want this to be widely known, but if so, here's a little blurb about it


kof:
what about this? https://github.com/cssinjs/react-jss#multi-tree-setup

jedwards1211:
Ah you're right, nevermind!

PR #192, opened 2018-01-01
Update babel setup

davidbonnet:
# Motivation
Old Babel presets cause warnings.

# Changes
* Remove deprecated presets
* Use `env` preset
* ~~Drop support for IE9 (discussion) by removing babel transform ES3 plugins~~

# TODO
* Apply changes on other cssinjs projects


.babelrc
```diff
@@ -1,9 +1,5 @@
 {
-  "presets": ["es2015", "react", "stage-0"],
-  "plugins": [
-    "transform-es3-member-expression-literals",
```
kof: what about those plugins?

davidbonnet: Whoops, I oversaw https://github.com/cssinjs/jss/issues/219#issuecomment-250696510 that mentions ES3 to support IE9. Since [Microsoft stopped supporting](https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support) this browser, should we still do so?

kof: we have to, because  react still does

davidbonnet: Missed [that](https://reactjs.org/blog/2016/01/12/discontinuing-ie8-support.html) as well.

kof:
sorry for letting this wait for so long, can you please update the pr, it go a conflict

kof:
we got a conflict

caub:
I updated babel a bit more in #241 (even if @babel is still marked beta-xx it's really ready

HenriBeck:
Solved by moving repo to new monorepo

PR #207, opened 2018-03-12
   Added plugin support for jss-rtl by default

yash-bagadia:
https://github.com/alitaheri/jss-rtl

        Enabling and disabling a react-jss plugin at runtime is too complex.
        So for supporting runtime toggle of rtl, added a flip option to sheet.

        if the jss-rtl plugin is enabled, then it checks for flip property in sheetOptions.
        So added a boolean property in sheet options before calling createStyleSheet.

        If its true, styles will be flipped.

        for toggling rtl, theme.rtl will have to be changed and in turn theme object
        will also have to be changed. So re render will happend and toggling will
        happen easily.


kof:
You should be doing this on a different layer, react-jss doesn't know anything about rtl. For e.g. if you update a theme, it will rerender everything.

PR #208, opened 2018-03-13
Setting new context to instance of createHoc in componentWillReceiveProps

yash-bagadia:
In the code for updateClassInstance method in react-dom.development.js file, first
all the lifecycle methods like componentWillReceiveProps, componentWillUpdate,
shouldComponentUpdate are called and in the end of updateClassInstance,
instance.context is set to new context.. so problem here is that in createHoc's
componentWillReceiveProps createState is called which in turn references this.context.
so this will always reference the old context. Since componentWillReceiveProps is
the first method in update cycle.. setting the context in this instance to newContext.

### DIFF
--- a/src/injectSheet.test.js
+++ b/src/injectSheet.test.js
@@ -282,9 +282,8 @@ describe('injectSheet', () => {
       const ThemeA = {color: "blue"};
       const ThemeB = {color: "red"};
       const ComponentA = injectSheet(() => ({a: {left: 2}}))()
-      const localJss = create()
       render((
-        <JssProvider jss={localJss}>
+        <JssProvider jss={jss}>
           <ThemeProvider theme={ThemeA}>
             <div>
               <ComponentA />
@@ -299,7 +298,7 @@ describe('injectSheet', () => {
       const actual = styleTags.map(innerText).map(trim).join('\n')
   
       expect(actual).to.be(stripIndent`
-      .NoRenderer-a-1-1 {
+      .a-id {
         left: 2;
       }
       `)
@@ -312,12 +311,7 @@ describe('injectSheet', () => {
           }
         }
       }
-      const newJss = createJss({
-        createGenerateClassName: () => {
-          let counter = 0
-          return rule => `${rule.key}-${counter++}`
-        }
-      })
+      const newJss = createJss({createGenerateClassName})
       newJss.use(rtl())
       render((
         <JssProvider jss={newJss}>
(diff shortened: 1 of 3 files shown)


kof:
lgtm, looking into the build fail

kof:
Can you please add a test for this?

yash-bagadia:
I will add a test case.. thanks..

kof:
Btw. it would be great to have a nice RTL demo on codesandox so we can link it from the documentation!

yash-bagadia:
Sure.. once this is merged will create a demo on sandbox.. still working on test case...

yash-bagadia:
@kof can you please help me to figure out why test is failing... is rendering not happening properly.. or rtl function is wrong? second expect is failing..still giving old value.

yash-bagadia:
correced the tests

src/tests/theming.js
```diff
@@ -303,6 +303,65 @@ describe('theming', () => {
       }
     `)
   })
+  it('setState should use new context', () => {
```
kof: the fix is not really about theming, it is about context used in the HOC, we should move the test to injectSheet.test.js

src/tests/theming.js
```diff
@@ -303,6 +303,65 @@ describe('theming', () => {
...
+    const ComponentA = injectSheet(() => ({a: {left: 2}}))()
+    render((
+      <JssProvider jss={localJss}>
+        <ThemeProvider theme={ThemeA}>
```
kof: do you really need a theme provider to reproduce it?

src/tests/theming.js
```diff
@@ -303,6 +303,65 @@ describe('theming', () => {
...
+        onProcessStyle(style, rule, sheet) {
+          return {
+            right: '2px'
+          }
+        }
+      }
+    }
+    const newJss = createJss({
+      ...preset(),
+      createGenerateClassName: () => {
```
kof: you can probably use the createGenerateClassName function from the tests/helper.js

src/tests/theming.js
```diff
@@ -303,6 +303,65 @@ describe('theming', () => {
...
+      return {
+        onProcessStyle(style, rule, sheet) {
+          return {
+            right: '2px'
+          }
+        }
+      }
+    }
+    const newJss = createJss({
+      ...preset(),
```
kof: also you don't really need the default preset here

yash-bagadia: Done.

src/injectSheet.test.js
```diff
@@ -306,9 +299,9 @@ describe('injectSheet', () => {
       const actual = styleTags.map(innerText).map(trim).join('\n')
   
       expect(actual).to.be(stripIndent`
-        .a-0 {
-          left: 2px;
-        }
+      .NoRenderer-a-1-1 {
```
kof: https://github.com/cssinjs/react-jss/blob/master/tests/helper.js#L1 this helper simplifies this

yash-bagadia: Done

PR #213, opened 2018-03-17
enhancing inject option to inject only part of theme as prop

yash-bagadia:
By default theme is injected as prop to component. But in the options to injectSheet
we can whitelist props to be injected with 'inject' property. However, if we want to inject only some part
of theme as prop there is no way. So adding one more key to inject option to give
the flexibility to inject only part of theme. 
So lets take an example component where we are consuming a property 'height' and we want to provide flexibility where property can directly be passed to the component or at a global level as part of theme so that all component instances will get the same value for that property. 
```javascript
const Example = props => {
const height = props.height || props.theme.ButtonBase.height; //give priority to props.height
//as it is passed specific to this instance
  return <div> value of height is {height} </div>;
};
```
While this works, there are two problems with this approach: 

1. Entire theme object is passed to the component which might be huge and the component required
only few of the properties.
2. For each of the property that we are allowing to be defined in theme, we will end up writing code
like:
```javascript
props.height || props.theme.ButtonBase.height
```
which makes the code look ugly and difficult to maintain.

By providing the provision to inject only part of theme as prop to the component, we solve
both of the above problems. Only required properties will be injected making props
object small.
Secondly, for each property  we just need to consume it like:
```javascript
props.height
```
and createHoc takes care of overriding theme passed prop if prop is passed to component
instance and it makes the code look independent of theme and easy to maintain.
One more advantage of extending inject option to provide what property to inject still
keeps react-jss code unaware of theme object and independent of it.

With this change we extend inject option to accept an object as well as one of the elements in the array with key 'themeProps' and its value will be the property path of the part required to be injected in the component as prop.

**Sample Usage:**
```javascript
const styles = {
color: 'red'
}
const Example = props => {
  const height = props.height;
  return <div> value of height is {height} </div>;
};
export injectSheet(styles, {inject: ['classes', {themeProps: 'ButtonBase'}]})(Example)
//and if we render this Example component wrapped with ThemeProvider:
<ThemeProvider theme={{color: 'blue', ButtonBase: {height: 10, width:20}}}><Example /> </ThemeProvider>
//height and width will be injected to props of Example
//Output will be: value of height is 10
```
Property passed to component instance will override property passed by theme:
```javascript
export injectSheet(styles, {inject: ['classes', {themeProps: 'ButtonBase'}]})(Example)
<ThemeProvider theme={{color: 'blue', ButtonBase: {height: 10, width:20}}}><Example height=30/> </ThemeProvider>
//width and height will be injected to props of Example
//output will be: value of height is 20
```
We can also pass path of a nested property in theme object to be injected as prop:
```javascript
export injectSheet(styles, {inject: ['classes', {themeProps: 'a.b.c'}]})(Example)
//so if theme object look like:
  theme = {
    color: 'red',
    a: {
      b: {
        c: {
          p: 10,
          q: 20
        }
      }
    }
  }
//p and q will be injected to props of Example.
```

### DIFF
--- a/src/createHoc.js
+++ b/src/createHoc.js
@@ -40,10 +40,22 @@ const getStyles = (stylesOrCreator, theme) => {
 
 // Returns an object with array property as a key and true as a value.
 const toMap = arr => arr.reduce((map, prop) => {
-  map[prop] = true
+  if (typeof prop === 'object' && prop.themeProps) {
+    map.themeProps = prop.themeProps
+  }
+  else {
+    map[prop] = true
+  }
   return map
 }, {})
-
+const getRequiredThemeProps = (object, property) => {
+  if (object && typeof object == 'object') {
+    if (typeof property == 'string' && property !== '') {
+      const split = property.split('.')
+      return split.reduce((obj, prop) => obj && obj[prop], object)
+    }
+  }
+}
 const defaultInjectProps = {
   sheet: false,
   classes: true,
@@ -214,6 +226,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       const props = {}
       if (injectMap.sheet) props.sheet = sheet
       if (isThemingEnabled && injectMap.theme) props.theme = theme
+      if (isThemingEnabled && injectMap.themeProps) {
+        const ThemeProps = getRequiredThemeProps(theme, injectMap.themeProps)
+        if (typeof ThemeProps === 'object') {
+          Object.assign(props, ThemeProps)
+        }
+      }
       Object.assign(props, this.props)
       // We have merged classes already.
       if (injectMap.classes) props.classes = classes
--- a/src/injectSheet.test.js
+++ b/src/injectSheet.test.js
@@ -70,7 +70,7 @@ describe('injectSheet', () => {
   })
 
   describe('injectSheet() option "inject"', () => {
-    const getInjected = (options) => {
+    const getInjected = (options, theme = {}, props = {}, bReturnPropsObject = false) => {
       let injectedProps
       const Renderer = (props) => {
         injectedProps = props
@@ -79,8 +79,8 @@ describe('injectSheet', () => {
       const MyComponent = injectSheet(() => ({
         button: {color: 'red'}
       }), options)(Renderer)
-      render(<ThemeProvider theme={{}}><MyComponent /></ThemeProvider>, node)
-      return Object.keys(injectedProps)
+      render(<ThemeProvider theme={theme}><MyComponent {...props} /></ThemeProvider>, node)
+      return bReturnPropsObject ? injectedProps : Object.keys(injectedProps)
     }
 
     it('should inject all by default', () => {
@@ -101,6 +101,16 @@ describe('injectSheet', () => {
 
     it('should inject classes and theme', () => {
       expect(getInjected({inject: ['classes', 'theme']})).to.eql(['theme', 'classes'])
+    })
+    it('should inject themeProps', () => {
+      expect(getInjected({inject: ['classes', {themeProps: 'ButtonBase'}]}, {color: 'red', ButtonBase: {height: 10}})).to.eql(['height', 'classes'])
+    })
+    it('should not break and inject classes', () => {
+      expect(getInjected({inject: ['classes', {themeProps: 'Button'}]}, {color: 'red', ButtonBase: {height: 10}})).to.eql(['classes'])
+    })
+    it('should override themeProps with prop passed to component', () => {
+      const heightProp = getInjected({inject: ['classes', {themeProps: 'ButtonBase'}]}, {color: 'red', ButtonBase: {height: 10}}, {height: 20}, true).height
+      expect(heightProp).to.eql(20)
     })
   })
 


yash-bagadia:
@kof : please go through the description of pull request and see it makes sense to you. This seems to be useful to me and it also maintains the low level behavior of react-jss as it still makes no assumptions about theme object.

kof:
So you are trying to port the same feature from mui. The difference though is that mui is a limited set of components, user is supposed to define a theme globally and defining default props in the same step over the theme can be easily done from the user perspective.

Here, given the fact that the components are unknown, you can't use some component name over the theme and then automatically apply. For e.g. there could be 3 different "Button" components in your application tree and this is out of control.

kof:
cc @oliviertassinari

yash-bagadia:
@kof agreed..but this change is intended towards components wrapped with injectSheet... for example in my Button component.. I just want the application which is consuming my button to know that if you want to set certain properties for all instances of button then set the properties under theme.ButtonBase (or theme.x.y.z or any path). So then its the responsibility of the application which builds the theme object if they want take advantage of this flexibility of setting properties of all instances at a global level in theme. react-jss or the component itself need not worry about it I guess.

kof:
you can inject 3 different Button components with injectSheet, you are not targeting an instance, you are literally targeting any Button component

yash-bagadia:
what I am telling is lets say my ComponentA says ... if you want to set props at global level for all instances... then you set props at path x.y in theme object... and I will call injectSheet like:
```javascript
injectSheet(styles, inject: ['theme', {themeProps: "x.y"}])
```
not dependent on component name at all.. just saying that for my component: you set the property at "x.y".. that is it... then you will be able to set for all instances of my component...

yash-bagadia:
also.. i guess its the responsibility of the library using react-jss to not have name conflicts if they want to take advantage of this.. .from react-jss point of view.. this is just an option if some component wants only part of theme.. forget props.. may be i just want to play with lets say palette in the theme.. so i can say give me only palette.. i dont want the entire theme object...

yash-bagadia:
i think react-jss need not worry about all this...

yash-bagadia:
So as an application which is using react-jss it is my responsibility for how many components I want to use this feature.. and if yes I want to use it, then it's my responsibility to define unique path in the theme object.

kof:
Ok, so you are not expecting react-jss to merge theme props based on a component name, you want to component provide a path, which react-jss uses then to merge in. I got it wrong. 

In that case it is very different from what MUI did recently. I am not sure there is a high demand on this feature.

kof:
Lets see if we get some 👍 on this, if yes, we can consider implementing it.

yash-bagadia:
Yes.. this is exactly what I meant... because as you said, react-jss works at a very low level.. i tried to make it independent of the theme object.. so component will define what properties I need from the theme object.. react-jss will just give those properties dumbly...

yash-bagadia:
ah.. In my application it's quite useful atleast because I am not using any custom styling/theming option... entirely we are relying on modules exposed by react-jss.. so it will be nice to have this without actually having to maintain my own version of injectSheet like material-ui..

kof:
Thoughts on the interface (independent of if we decide to add this or not):

1. `['classes', {themeProps: 'ButtonBase'}]` this is not very nice, it needs something better
1. instead of "x.y" as a string, it is evtl. easier to provide a fn: `(theme) => theme.x.y`

yash-bagadia:
@kof so this option has to be passed to injectSheet which is kind of a static composition.. not sure if we can achieve dynamic behavior here like fn: ``` (props) => props.theme.x.y ```...
am i missing something here?

kof:
What is the difference to `x.y` as a string? if theme changes you have to update.

yash-bagadia:
so options are passed to injectSheet and on the basis of that we are preparing props in createHoc.. so how at that time how will props have theme? 
Also theme is anyways part of state of createHoc.. so will get new props automatically

yash-bagadia:
Since this is a PR, few people will look at it.. to see if people need it I will create a issue out of it..

oliviertassinari:
As I understand this change, it's something that people can already do on their side. Moving the logic to *react-jss* provides limited value. The approach on Material-UI side is a bit different. We are more biased. People can't choose the path they want. The pros are over predictability and minimized overhead.

kof:
Ro do that user would write a utility function, pass to it props and get props back and that function would check if there are any theme properties which can act as default props in case default props are missing. This is totally possible inside of  a user component, @yash-bagadia wants to take this outside of the component itself into injectSheet call for "better" dx.

I am not sold on this because I almost never needed it, if I did, using var something = props.something || theme.something was not too bad.

kof:
I would recommend for now using a utility like this:

```js
const mergeProps = (props, get) => {
  const {theme} = props
  if (!theme) return props
  return {...get(theme), ...props}
}
```
then in userland

```js
mergedProps = mergeProps(props, (theme) =>  ({theme.something})) 
```

kof:
Actually I think what you want is props reducer function, which @kentcdodds was talking lately about

kentcdodds:
Oh, you mean this old thing?

https://blog.kentcdodds.com/the-state-reducer-pattern-b40316cfac57

:)

kof:
@kentcdodds yep, thanks for the link. You used it as a direct api on a component to manage its state. I think it is same good to use this pattern when you generate a HOC in order to pass the props through that reducer which then go to the child component. It gives user a power to actually decide what props go to the child component.

kof:
```js
// whenever props has a color - it will be used, otherwise theme.color will be used
const reduceProps = props => ({color: props.theme.color, ...props})
injectSheet(styles)(Button, {reduceProps})
```

kof:
Actually with this I could have avoided as well the built-in "inject" option, user could implement any logic using that function, it could be even a static one for all injectSheet calls:

```js
// Custom global props reducer
injectSheet.reduceProps = (props) => props
```

kof:
Current props injection behaviour could be implemented like this:
```js
injectSheet.reduceProps = ({sheet, theme, classes, ...rest}) => ({
  ...rest,
  theme,
  classes
})
```

In that case we remove sheet from the props, but keep the theme and classes. We could as well do something else for e.g. let props overwrite classes prop by inverting the order and remove the theme prop as well

```js
injectSheet.reduceProps = ({sheet, theme, classes, ...rest}) => ({
  classes,
  ...rest
})
```

yash-bagadia:
yes @kof I spent some time and came up with my own withStyles to achieve this:

```javascript
// tslint:disable-next-line
const hoistNonReactStatics = require("hoist-non-react-statics");
const getThemeProps = (theme: any, property: string = "") => {
  if (property !== "") {
    const split = property.split(".");
    return split.reduce((obj, prop) => {
      return obj && obj[prop];
    }, theme);
  }
};
const withStyles = (styles: any, WrappedComponent: any, property?: string) => {
  const EnhancedComponent = (props: any) => {
    const { theme, ...otherProps } = props;
    const ThemeProps = getThemeProps(theme, property);
    return <WrappedComponent {...ThemeProps} {...otherProps} />;
  };
  return injectSheet(styles)(
    hoistNonReactStatics(EnhancedComponent, WrappedComponent)
  );
};
```

kof:
@yash-bagadia I guess you don't need hoistNonReactStatics if you use injectSheet from react-jss, but other than that if you want your syntax in exact this way its ok to do that.

kof:
Oh just realized you are introducing an additional container by this. I you could avoid this if we implement the new reducer api mentioned above.

yash-bagadia:
@kof yeah.. the propsReducer is a much cleaner solution to this problem 👍 may be I could try this in the PR?

yash-bagadia:
closing this as reimplemented this with: https://github.com/cssinjs/react-jss/pull/216/files

PR #216, opened 2018-03-23
 Introducing reduceProps functionality replacing inject option

yash-bagadia:
react-jss had one option called 'inject' which was used to whitelist properties injected by injectSheet to the child component:

```javascript
// Only `classes` prop will be passed by the ReactJSS HOC now. No `sheet` or `theme`.
const Button = injectSheet(styles, {inject: ['classes', 'sheet']})(
  ({classes}) => <button>My button</button>
)
```
However, if we want to inject only some part
of theme as prop there is no way. So adding one more key to inject option to give
the flexibility to inject only part of theme. 
So lets take an example component where we are consuming a property 'height' and we want to provide flexibility where property can directly be passed to the component or at a global level as part of theme so that all component instances will get the same value for that property. 
```javascript
const Example = props => {
const height = props.height || props.theme.ButtonBase.height; //give priority to props.height
//as it is passed specific to this instance
  return <div> value of height is {height} </div>;
};
```
While this works, there are two problems with this approach: 

1. Entire theme object is passed to the component which might be huge and the component required
only few of the properties.
2. For each of the property that we are allowing to be defined in theme, we will end up writing code
like:
```javascript
props.height || props.theme.ButtonBase.height
```
which makes the code look ugly and difficult to maintain.

With this change, we replace inject option with another option called 'reduceProps' whose value should be a function that should return the props to be injected.

By providing the provision to inject only part of theme as prop to the component, we solve
both of the above problems. Only required properties will be injected making props
object small.
Secondly, for each property  we just need to consume it like:
```javascript
props.height
```
and createHoc takes care of overriding theme passed prop if prop is passed to component
instance and it makes the code look independent of theme and easy to maintain.

**Sample Usage**
```javascript
const styles = {
color: 'red'
}
const Example = props => {
  return <div className={props.classes.hello}> Hello </div>;
};
const reduceProps = (props) => {
   const {theme, classes, sheet, ...rest} = props
    return {theme, classes, ...rest}
 }
export injectSheet(styles, {reduceProps})(Example)
```
So, with this change user can write any function and decide what properties should go to the child component. Not just for whitelisting some properties like theme, classes, sheet. But we can even whitelist some properties within these objects:

```javascript
const styles = {
color: 'red'
}
const reduceProps = (props) => {
   const {theme, classes, sheet, ...rest} = props
    return {classes, ...theme.ButtonBase, ...rest}
 }
const Example = props => {
  const height = props.height;
  return <div> value of height is {height} </div>;
};
export injectSheet(styles, {reduceProps })(Example)
//and if we render this Example component wrapped with ThemeProvider:
<ThemeProvider theme={{color: 'blue', ButtonBase: {height: 10, width:20}}}><Example /> </ThemeProvider>
//height and width will be injected to props of Example
//Output will be: value of height is 10
```
reduceProps function can also be set a global level by passing a prop to the JssProvider:
```javascript
const styles = {
color: 'red'
}
//only thene and classes will be injected
const reduceProps = (props) => {
   const {theme, classes, sheet, ...rest} = props
    return {classes, ...theme.ButtonBase, ...rest}
 }
const Example = props => {
  const height = props.height;
  return <div> value of height is {height} </div>;
};
export injectSheet(styles)(Example)
//and if we render this Example component wrapped with JssProvider:
<JssProvider reduceProps={reduceProps}><ThemeProvider theme={{color: 'blue', ButtonBase: {height: 10, width:20}}}><Example /> </ThemeProvider></JssProvider>
//height and width will be injected to props of Example
//Output will be: value of height is 10
```

### DIFF
--- a/src/injectSheet.js
+++ b/src/injectSheet.js
@@ -29,11 +29,6 @@ export default function injectSheet(stylesOrSheet, options = {}) {
     options.index = indexCounter++
   }
   return (InnerComponent = NoRenderer) => {
-    const {reduceProps} = options
-    // if reduceProps is not in options, check if static property for the function is set
-    if (typeof reduceProps !== 'function' && typeof injectSheet.reduceProps === 'function') {
-      options.reduceProps = injectSheet.reduceProps
-    }
     const Jss = createHoc(stylesOrSheet, InnerComponent, options)
     return hoistNonReactStatics(Jss, InnerComponent, {inner: true})
   }
--- a/src/injectSheet.js
+++ b/src/injectSheet.js
@@ -29,6 +29,11 @@ export default function injectSheet(stylesOrSheet, options = {}) {
     options.index = indexCounter++
   }
   return (InnerComponent = NoRenderer) => {
+    const {reduceProps} = options
+    // if reduceProps is not in options, check if static property for the function is set
+    if (typeof reduceProps !== 'function' && typeof injectSheet.reduceProps === 'function') {
+      options.reduceProps = injectSheet.reduceProps
+    }
     const Jss = createHoc(stylesOrSheet, InnerComponent, options)
     return hoistNonReactStatics(Jss, InnerComponent, {inner: true})
   }
--- a/readme.md
+++ b/readme.md
@@ -386,18 +386,34 @@ const Button = injectSheet(buttonStyles)(() => <button><Label>my button</Label><
 ## Whitelist injected props
 
 By default "classes" and "theme" are going to be injected to the child component over props. Property `theme` is only passed when you use a function instead of styles object.
-If you want to whitelist some of them, you can now use option `inject`. For e.g. if you want to access the StyleSheet instance, you need to pass `{inject: ['sheet']}` and it will be available as `props.sheet`.
-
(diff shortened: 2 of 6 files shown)


src/injectSheet.js
```diff
@@ -29,6 +29,11 @@ export default function injectSheet(stylesOrSheet, options = {}) {
     options.index = indexCounter++
   }
   return (InnerComponent = NoRenderer) => {
+    const {reduceProps} = options
+    // if reduceProps is not in options, check if static property for the function is set
+    if (typeof reduceProps !== 'function' && typeof injectSheet.reduceProps === 'function') {
```
HenriBeck: I think it would be better to add a prop to the JssProvider rather than defining it globally on the injectSheet function.

yash-bagadia: @HenriBeck 👍  Done.

kof:
Sorry for taking so long, I will try to get back to this soon.

HenriBeck:
@yash-bagadia this repo was moved to https://github.com/cssinjs/jss/tree/master/packages/react-jss

Please recreate your PR if you want this feature.

Issue #212, opened 2018-03-14
Adding ref support for injectSheet hoc.


alexankin:
In our project we need to have access to original component after it has been wrapped with injectSheet hoc.

Example:
```jsx
const Styled = injectSheet(InnerComponent);
let component;
<Styled ref = {cmp => component = cmp}/>
```
In this case component will be injectSheet(Component), not Component. And this is expected behavior. But we need to have acess to InnerComponent, something like this:
```jsx
<Styled componentRef = {cmp => component = cmp}/>
```

TriPSs:
Can't you just set the `componentRef` yourself? Like:

```js
<Styled componentRef={ref => component = ref} />
```

```js
class Styled extends React.Component {

  render() {
    return (
       <div ref={this.props.componentRef}>

       </div
    )
  }
}
```

alexankin:
I need ref to InnerComponent itself, not to any html element inside it. So `ref={this.props.componentRef}` should be done in injectSheet hoc.

TriPSs:
Pharaps then update the render method of `createHoc` to

```js
  render() {
    const { theme, dynamicSheet, classes } = this.state
    const { ref = null, ...rest } = this.props
    
    const sheet = dynamicSheet || this.manager.get(theme)
    const props = {}
    
    if (injectMap.sheet) props.sheet = sheet
    if (isThemingEnabled && injectMap.theme) props.theme = theme
    
    Object.assign(props, rest)
    // We have merged classes already.
    if (injectMap.classes) props.classes = classes
    return <InnerComponent ref={ref} {...props} />
  }
```

Ideas @kof?

HenriBeck:
@TriPSs Your example wouldn't work as you don't get the `ref` property inside `this.props`. We would need some prop named `componentRef` for that.

You should also be able to call `createRef` for example in the constructor of the component you need access too.
That should work too:
```js
// In your component where you need access
constructor(props) {
  super(props);

  props.createRef(this);
}
```

TriPSs:
@HenriBeck Updated the example, forgot that ref would not be available in the props. Thanks!

kof:
I am open to merge a PR for this.

kof:
Normally other libraries call this prop `innerRef`

kof:
Please also include a test

alexkuz:
[...]

PR #219, opened 2018-03-30
Added support for innerRef

TriPSs:
Added support for innerRef, issue #212


src/createHoc.js
```diff
@@ -210,14 +214,20 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
 
     render() {
       const {theme, dynamicSheet, classes} = this.state
+      const {innerRef, ...rest} = this.props
+
       const sheet = dynamicSheet || this.manager.get(theme)
       const props = {}
```
kof: we can just rename ...rest to ...props and use it, so we don't need to use Object.assign later on as well

src/tests/inner-ref.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable global-require, react/prop-types */
+import expect from 'expect.js'
+import React, {PureComponent} from 'react'
+import {spy} from 'sinon'
+
+describe('inner ref', () => {
+  it('should provide a ref on the inner component', () => {
+    const handleRef = spy()
+
+    /* eslint-disable react/no-multi-comp, react/prefer-stateless-function */
```
kof: why not use a functional component?

src/createHoc.js
```diff
@@ -210,14 +214,20 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
...
+
       if (injectMap.sheet) props.sheet = sheet
```
kof: in that case you will have to add an additional condition here in order to not override

src/tests/inner-ref.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable global-require, react/prop-types */
+import expect from 'expect.js'
+import React, {PureComponent} from 'react'
+import {spy} from 'sinon'
+
+describe('inner ref', () => {
```
kof: this test belongs to injectSheet.test.js

TriPSs: Yes, i did not see that test file!

src/tests/inner-ref.js
TriPSs: Because of:
image

src/createHoc.js
TriPSs: Override what exactly?

kof: In other words we need to make sure that props passed by the user override hoc props. At least that was the logic so far.

TriPSs: Makes sense.

Think we have to do the same for theme?
```js
      if (injectMap.sheet && !props.sheet) props.sheet = sheet
      if (isThemingEnabled && injectMap.theme && !props.theme) props.theme = theme
```

kof: yep

HenriBeck:
@TriPSs can you add documentation for it?

kof:
I am already on it

kof:
Ok, released 8.4.0, also I realised that you mised up inner component and inner element in the tests. What react calls refs are references to component instances aka elements. References to components are normal javascript references.

kof:
in any case good work on landing this one @TriPSs and thanks @HenriBeck for looking over. Please test the release if something is wrong I am back in 9h.

HenriBeck:
https://reactjs.org/docs/forwarding-refs.html
Now that react 16.3 is released there is a supported way of forwarding refs.

kof:
Do we need any changes to support forwardRef?

HenriBeck:
Yes, we either need to release a version only for `react@^16.3` or we would need to check if `forwardRef` exists on the `React` object and only use it when it's actually supported.

If we were to release a separate version for `react@^16.3` we could also use the new lifecycle methods and deprecate the old ones maybe.

kof:
If we assume that `React. forwardRef`  will be used by the end consumer, it seems that we don't need to do anything, it will just work (from looking at the docs).

Issue #220, opened 2018-04-02
Use symbol where evailable


kof:
Inspired by react

https://github.com/facebook/react/blob/bc70441c8b3fa85338283af3eeb47b5d15e9dbfe/packages/shared/ReactSymbols.js

We currently use for the same purpose md5 strings, but we could first check if we can use symbols https://github.com/cssinjs/react-jss/blob/master/src/ns.js

kof:
wait, we still have no running tests, need to fix it, we can't release anything right now…

PR #221, opened 2018-04-02
Use symbols for namespacing

HenriBeck:
#220


src/ns.js
```diff
@@ -1,7 +1,17 @@
+const hasSymbol = typeof Symbol === 'function' && Symbol.for
+
 /**
  * Namespaces to avoid conflicts on the context.
  */
-export const jss = '64a55d578f856d258dc345b094a2a2b3'
-export const sheetsRegistry = 'd4bd0baacbc52bbd48bbb9eb24344ecd'
-export const managers = 'b768b78919504fba9de2c03545c5cd3a'
-export const sheetOptions = '6fc570d6bd61383819d0f9e7407c452d'
+export const jss = hasSymbol
```
kof: does this lint? I thought I disabled multiline ternaries in linter?

src/ns.js
```diff
@@ -1,7 +1,17 @@
+const hasSymbol = typeof Symbol === 'function' && Symbol.for
```
kof: I would also rename ns.js to symbols.js and other "ns" variables for the same thing

src/ns.js
HenriBeck: I didn't get any lint errors

kof:
uhh, just tried, all test fail …

kof:
need to fix the CI…

kof:
@HenriBeck do you know whats wrong with tests?

HenriBeck:
@kof I see this error:  No usable sandbox! Update your kernel or see https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md for more information on developing with the SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.

kof:
I mean locally, if you run `npm test` I get a bunch of errors

HenriBeck:
It has something to do with the symbols but I'm not quite sure what.

HenriBeck:
It works in the browser and does what it should do: https://codesandbox.io/s/0mn0vyp6wv

kof:
Yep, we still need to fix tests though :)

HenriBeck:
I have a free day tomorrow and will try to fix the tests then

kof:
cool, please try to fix tests with symbols first, they block releases, otherwise we need to remove them from master

PR #223, opened 2018-04-02
Added support for forwardRef API

HenriBeck:
This PR adds support for the forwardRef API introduced in React 16.3.
This will only use the new API when it's available and normal ref function is actually being used.

The `innerRef` prop will still work.

I will still need to write some tests and do some testing to make sure it works as intended and that the backward compatibility will work.

### DIFF
--- a/src/createHoc.js
+++ b/src/createHoc.js
@@ -74,7 +74,7 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
   const defaultProps = {...InnerComponent.defaultProps}
   delete defaultProps.classes
 
-  return class Jss extends Component {
+  class Jss extends Component {
     static displayName = `Jss(${displayName})`
     static InnerComponent = InnerComponent
     static contextTypes = {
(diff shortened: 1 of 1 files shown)


src/createHoc.js
```diff
@@ -229,11 +229,17 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
   }
 
   function forwardRef(props, ref) {
+    if (React.forwardRef && ref) {
```
kof: you mean forwardedRef

src/createHoc.js
```diff
@@ -229,11 +229,17 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
   }
 
   function forwardRef(props, ref) {
```
kof: I am wondering if we need it at all. Why not let user wrap it?

src/createHoc.js
kof: I thought the point of the React.forwardRef is to let user wrap any component and receive the ref of a component which is not a hoc…

src/createHoc.js
kof: Hmm, no, you are right, we need to wrap it

src/createHoc.js
```diff
@@ -227,4 +227,23 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
...
+        />
+      )
+    }
+
+    return (
+      <Jss {...props} />
+    )
+  }
+
+  forwardRef.displayName = `forwardRef(${displayName})`
```
kof: is forwardRef generating a hoc?

src/createHoc.js
```diff
@@ -227,4 +227,23 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       return <InnerComponent ref={innerRef} {...props} />
     }
   }
+
+  function forwardRef(props, ref) {
+    if (React.forwardRef && ref) {
```
kof: why are you checking for React.forwardRef here, you are not going in here when React.forwardRef is not defined

src/createHoc.js
```diff
@@ -227,4 +227,23 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
...
+    }
+
+    return (
+      <Jss {...props} />
+    )
+  }
+
+  forwardRef.displayName = `forwardRef(${displayName})`
+
+  return React.forwardRef ? React.forwardRef(forwardRef) : Jss
```
kof: it seems like all we need it:
```js

if (React.forwardRef) {
  return React.forwardRef((props, ref) => <Jss {...props} innerRef={ref} />)
}

return Jss

File: src/createHoc.js (hunk 3 above)
Comment (HenriBeck) [reply src/createHoc.js]: `forwardRef` is not generating a HOC, it's actually generating a special React Node Type. But you can set the `displayName` so it shows it in the React DevTools not just as `ForwardRef`.

**Comment by HenriBeck** [comment]
Okay, so I tested it.

When the `forwardRef` API is available it will support it, but without deprecating the `innerRef` prop.
Also when the `forwardRef` API is being used, the Node in the ReactDevTools will show up as `ForwardRef(Jss(Component))`.

When the `forwardRef` API is not available it will just work as normal.

File: src/createHoc.js [hunk 6]
```diff
@@ -93,25 +93,51 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       this.state = this.createState({theme}, props)
     }
 
-    get jss() {
-      return this.context[ns.jss] || optionsJss || jss
+    componentWillMount() {
```
kof: many changes here with changed order of the methods, makes it hard to review, was there a particular reason? It is better to focus on specific changes per PR to get the review faster.

HenriBeck: The reason for the ordering was that I was getting an ESLint issue. Apparently the sort-comp rule doesn't work properly when the class is returned

kof:
Ok, should we add a test for this?

kof:
Do we need to add anything to the docs?

kof:
Lets continue here https://github.com/cssinjs/react-jss/pull/228

PR #234, opened 2018-04-16
Improving the README documentation for clarity and better examples

nai888:
Fixed some small grammar mistakes. Fixed an error in the Table of Contents and headers. Added yarn for install. Simplified and expanded the Basic example to show a full working document, and provided the resulting code for clarity. Added more variables to the Dynamic Values example, including showing passing a value as a prop for clarity, and provided the resulting code for clarity.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -3,7 +3,7 @@
 [![Gitter](https://badges.gitter.im/JoinChat.svg)](https://gitter.im/cssinjs/lobby)
 [![Build Status](https://travis-ci.org/cssinjs/react-kss.svg?branch=master)](https://travis-ci.org/cssinjs/react-jss)
 
-React-JSS provides components for [JSS](https://github.com/cssinjs/jss) as a layer of abstraction. JSS and [presets](https://github.com/cssinjs/jss-preset-default) are already built in! Try it out in the [playground](https://codesandbox.io/s/j3l06yyqpw).
+React-JSS provides components for [JSS](https://github.com/cssinjs/jss) as a layer of abstraction. JSS and the [default preset](https://github.com/cssinjs/jss-preset-default) are already built in! Try it out in the [playground](https://codesandbox.io/s/j3l06yyqpw).
 
 Benefits compared to lower level core:
 
@@ -19,6 +19,7 @@ Benefits compared to lower level core:
 * [Install](#install)
 * [Usage](#usage)
   * [Basic](#basic)
+  * [Dynamic Values](#dynamic-values)
   * [Theming](#theming)
   * [Server-side rendering](#server-side-rendering)
   * [Reuse styles in different components](#reuse-styles-in-different-components)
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -35,68 +36,156 @@ Benefits compared to lower level core:
...
 }
 
-const Button = ({classes, children}) => (
-  <button className={classes.button}>
-    <span className={classes.label}>
+// Define the component using these styles and pass it the 'classes' prop.
+// Use this to assign scoped class names.
+const Button = ({ classes, children }) => (
+  <button className={classes.myButton}>
+    <span className='myLabel'>
```
kof: I think we should not advise on using global class names, as we are trying to fix this.

readme.md
```diff
@@ -35,68 +36,156 @@ Benefits compared to lower level core:
...
 import React from 'react'
+import { render } from 'react-dom'
+// Import React-JSS
 import injectSheet from 'react-jss'
 
+// Create your Styles. Remember, since React-JSS uses the default preset,
+// most plugins are available without further configuration needed.
 const styles = {
-  button: {
-    background: props => props.color
```
kof: why not to keep function value example as well?

readme.md
```diff
@@ -35,68 +36,156 @@ Benefits compared to lower level core:
...
+    <span className={classes.myLabel}>
       {children}
     </span>
   </button>
 )
 
 Button.defaultProps = {
-  fontWeight: 'bold',
-  color: 'red'
+  pad: 10,
```
kof: why not use full names, like padding?

readme.md
nai888: Fair enough. I was just trying to make it clearer to people brand-new to CSSinJS what would happen if you use a global variable like that, like you do in your docs for [jss-nested](http://cssinjs.org/jss-nested?v=v6.0.1#use--to-reference-selector-of-the-parent-rule), but that's a fair point about not wanting to encourage it.

readme.md
nai888: When I was reading the docs the first time, it was far more confusing than it was helpful because I had no idea what it was doing yet. I think it's really valuable to have one basic example that's easily grasped by someone completely new to it, and then incrementally add more complexity.

Plus, since function values are explained in the very next section, I don't think anything's lost by removing it.

readme.md
nai888: I was trying to make it clearer that the props used in styling don't *have* to match the name of the CSS rule, but perhaps that's not an important distinction (and is perhaps obvious to most React users). I can change it back to the full name, no problem.

readme.md
kof: true

readme.md
kof: ok, then lets use some other word like spacing

kof:
lgtm, thanks!

PR #236, opened 2018-04-16
Revert symbols

### LINKED ISSUE #229: Fix tests broken by symbols

HenriBeck:
Apparently, the old context of React doesn't like symbols as keys.
Fixes #229


kof:
merged, thanks!

PR #241, opened 2018-05-02
update most dev deps

caub:
not finished with eslint

depends on https://github.com/cssinjs/eslint-config-jss/pull/2

**Question**: are you fine if I switch from mocha + expect.js to jest? it'll be very similar and a bit simpler (I'd do it on jss as well)?


src/JssProvider.test.js
```diff
@@ -276,14 +276,12 @@ describe('JssProvider', () => {
...
 
-      renderToString(
-        <JssProvider registry={customSheets} jss={localJss}>
-          <div>
-            <ComponentA />
-            <ComponentB />
-          </div>
-        </JssProvider>
-      )
+      renderToString(<JssProvider registry={customSheets} jss={localJss}>
```
caub: I don't get it, but eslint --fix is forcing it like this..

caub:
reverted from babel7 to babel6, safer

kof:
I am fine with jest if you are ready to refactor all tests

.eslintrc
```diff
@@ -2,6 +2,7 @@
   "parser": "babel-eslint",
   "extends": "eslint-config-jss",
   "globals": {
+    "window": true,
```
kof: can't imagine it is not already part of the eslint-config-jss which extends airbnb

caub: exact, thanks, fixed

Ok I'll look into mocha->jest, maybe not easy with karma usually jest uses jsdom

HenriBeck:
Solved by moving this repo to the new monorepo.

PR #246, opened 2018-05-21
Update dependency to 2.5.0

sergivillar:
I want to update 'hoist-non-react-statics' dependency in order to allow working with React 16.3 version and new lifecycle methods.


kof:
merged, thanks

Issue #167, opened 2017-10-22
If there is no registry ssr, don't render styles


kof:
See original discussion https://github.com/callemall/material-ui/pull/8757

Problem: if we disable rendering without registry, how will user discover whats wrong?

- do we need  explicit prop?
- should we warn when ssr and no registry?

kof:
released 8.6.0

kof:
thanks @in19farkt

PR #265, opened 2018-06-24
Added disabling of styles generation

in19farkt:
For traversing the React tree outside of the HTML rendering, you should add `disableStylesGeneration` property.

```javascript
import * as React from 'react';
import * as ReactDOM from 'react-dom';
import bootstrapper from 'react-async-bootstrapper';

import { JssProvider } from 'react-jss';
import MyApp from './MyApp';

const App = ({ disableStylesGeneration }) => (
  <JssProvider disableStylesGeneration>
    <MyApp />
  </JssProvider>
);

async function main() {
  await bootstrapper(<App disableStylesGeneration />);
  ReactDOM.render(<App />, document.getElementById('root'));
}

main();

```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -22,6 +22,7 @@ Benefits compared to lower level core:
   * [Dynamic Values](#dynamic-values)
   * [Theming](#theming)
   * [Server-side rendering](#server-side-rendering)
+  * [React tree traversing](#react-tree-traversing)
   * [Reuse styles in different components](#reuse-styles-in-different-components)
   * [The inner component](#the-inner-component)
   * [The inner ref](#the-inner-ref)
@@ -326,6 +327,33 @@ export default function render(req, res) {
     </html>
   ))
 }
+```
+
+### React tree traversing
+
+For traversing the React tree outside of the HTML rendering, you should add `disableStylesGeneration` property.
+
+```javascript
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import bootstrapper from 'react-async-bootstrapper';
+
+import { JssProvider } from 'react-jss';
+import MyApp from './MyApp';
+
+const App = ({ disableStylesGeneration }) => (
+  <JssProvider disableStylesGeneration>
+    <MyApp />
+  </JssProvider>
+);
+
+async function main() {
+  await bootstrapper(<App disableStylesGeneration />);
+  ReactDOM.render(<App />, document.getElementById('root'));
+}
+
+main();
+
 ```
 
 ### Reuse styles in different components
--- a/src/JssProvider.js
+++ b/src/JssProvider.js
@@ -1,5 +1,5 @@
 import {Component, Children} from 'react'
-import {node, func, string} from 'prop-types'
+import {node, func, string, bool} from 'prop-types'
 import {createGenerateClassNameDefault} from './jss'
 import * as ns from './ns'
 import contextTypes from './contextTypes'
@@ -10,6 +10,7 @@ export default class JssProvider extends Component {
     ...propTypes,
     generateClassName: func,
     classNamePrefix: string,
+    disableStylesGeneration: bool,
     children: node.isRequired
   }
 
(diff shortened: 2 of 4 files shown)


kof:
https://github.com/cssinjs/react-jss/issues/167

I was thinking to just use missing registry as a sign to not generate the styles, wdyt?

in19farkt:
I think this behavior is not obvious.

kof:
Why? On ssr, the only thing that get the rendered styles is a registry. If it is missing, noone is going to use those styles, why generating them …

in19farkt:
And what to do on the client's side?

kof:
Nothing, on the client side we don't have a need to render components without styles? Or do we?

in19farkt:
Yes, but if we are on the client side will try to traverse the React tree, we generate classes that are not in the layout which came from the server. And `ReactDOM.hydrate` will not work.

kof:
How is that possible? Styles generated on the server should always correspond to those on the client.

in19farkt:
We have asynchronous features, which are always on the server, but they are loaded on the client as needed. Which leads to a different order of rendering. To avoid this, we need to first load all the necessary features without generating styles, and only after that, we can render a tree identical to the server one.

kof:
Let me understand this clearly:
So you render a component A and component B on the server. Then on the client you rehydrate, but component B is not part of the rehydration somehow, because it is not needed yet?

in19farkt:
Not exactly. I will make a detailed example to understand the essence of our problem.

kof:
I appreciate that, understanding the use cases very well is the core of making good api decisions.

in19farkt:
For example, we have something layout:
```javascript
const SomeLayout = () => (
  <App>
    <AsyncFeatureA />
    <ComponentB />
    <AsyncFeatureC />
    <ComponentD />
  </App>
)

const AsyncFeatureC = ({ classes }) => (
  <div className={classes.root}>
    <ContentC />
  </div>
)
```
## On server:
```javascript
renderToString(<SomeLayout />)
```
html
```html
<div class="App-root-1">
  <div class="AsyncFeatureA-root-2" />
  <div class="ComponentB-root-3" />
  <div class="AsyncFeatureC-root-4">
    <div class="ContentC-root-5">
  </div>
  <div class="ComponentD-root-6" />
</div>
```
styles
```css
.App-root-1 {} // Unthemed, Static
.AsyncFeatureA-root-2 {}
.ComponentB-root-3 {}
.AsyncFeatureC-root-4 {}
.ContentC-root-5 {}
.ComponentD-root-6 {}
```
## On client:
```javascript
// tree traversing for load async features
await bootstrapper(<SomeLayout />); // first styles injecting
ReactDOM.hydrate(<SomeLayout />, document.getElementById('root')); // second styles injecting
```
First styles injecting:
```css
.App-root-1 {} // Unthemed, Static
.ComponentB-root-2 {}
.ComponentD-root-3 {}
```
In the case that there are static and unthemed styles, the second generation of styles leads to a shift in the numbering of classes. And we'll get something like this after the rendering:
html:
```html
<div class="App-root-1">
  <div class="AsyncFeatureA-root-1" />
  <div class="ComponentB-root-2" />
  <div class="AsyncFeatureC-root-3">
    <div class="ContentC-root-4">
  </div>
  <div class="ComponentD-root-5" />
</div>
```
Second styles injecting:
```css
.AsyncFeatureA-root-1 {}
.ComponentB-root-2 {}
.AsyncFeatureC-root-3 {}
.ContentC-root-4 {}
.ComponentD-root-5 {}
```
Result styles in head:
```css
.App-root-1 {} // Unthemed, Static
.ComponentB-root-2 {}
.ComponentD-root-3 {}
.AsyncFeatureA-root-1 {}
.ComponentB-root-2 {}
.AsyncFeatureC-root-3 {}
.ContentC-root-4 {}
.ComponentD-root-5 {}
```

in19farkt:
This solution avoids such problems. I think we should be able to turn off generation and injection styles if we need it.

kof:
I am missing the point, why would we need that on the client? If server-side rendering works correctly and you generate styles only once, why wouldn't it work on the client later on with the same class names?

kof:
Is bootstrapper something you run on the client as well?

in19farkt:
Bootstrap function traverses a react tree for some additional logic connected with async features loading. It analyzes some things to do it correctly. So when we traverse a tree, we just need to collect some info, but now traversing also starts styles injecting before the moment, when it really needs.

Maybe you have other solution for how can i inspect react tree without running styles injection?

in19farkt:
The tree is traversed using the [react-tree-walker](https://github.com/ctrlplusb/react-tree-walker) library.

kof:
Do you use bootstrapper on the client or on the server or on both?

in19farkt:
On both

kof:
Ok, that makes it clear now.

kof:
Ok I agree with the public API now, lets go into implementation details

kof:
actually your implementation  is fine, merging it, thanks!

PR #266, opened 2018-07-02
Update #123 Hot Module Replacement fix to current version, and improve unit test

felthy:
I’d love to get hot module replacement working for JSS styles in my projects so I’ve had a go at bringing pull request #123 up to date. What I've done is:

* merge `master` into [`fix/hmr`](https://github.com/felthy/react-jss/tree/fix/hmr)
* as part of the merge, change the `isHmrUpdate` check because `this.manager.keys.length === 0` is not true by the time `componentWillUpdate()` is called during hot replacement
* move the unit test to `src/tests/hot-reloading.js` and adjust it to use [react-hot-loader](https://github.com/gaearon/react-hot-loader)’s `register()` function as suggested by [@patrickml](https://github.com/cssinjs/react-jss/pull/123#issuecomment-317517819)

With these changes, HMR is now working for me with the following JSS setup (adapted from [create-react-app](https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/template/src/App.js)):

```javascript
import React, { Component } from 'react'
import PropTypes from 'prop-types'
import {compose} from 'recompose'
import logo from './logo.svg'
import {hot} from 'react-hot-loader'

class App extends Component {
  static propTypes = {
    classes: PropTypes.object.isRequired
  }

  render () {
    const {classes} = this.props
    return (
      <div className={classes.app}>
        <header className={classes.header}>
          <img src={logo} className={classes.logo} alt='logo' />
          <h1 className={classes.title}>Welcome to React</h1>
        </header>
        <p className={classes.intro}>
          To get started, edit <code>src/App.js</code> and save to reload.
        </p>
      </div>
    )
  }
}

export default compose(
  hot(module),

  injectSheet({
    app: {
      textAlign: 'center'
    },

    logo: {
      animation: 'App-logo-spin infinite 20s linear',
      height: 80
    },

    header: {
      backgroundColor: '#22f',
      height: 150,
      padding: 20,
      color: 'white'
    },

    title: {
      fontSize: '1.5em'
    },

    intro: {
      fontSize: 'large'
    },

    '@keyframes App-logo-spin': {
      from: { transform: 'rotate(0deg)' },
      to: { transform: 'rotate(360deg)' }
    }
  })
)(App)
```


felthy:
I’ve also published this version at [@felthy/react-jss](https://www.npmjs.com/package/@felthy/react-jss) if anyone wants to give it a try.

src/createHoc.js
```diff
@@ -109,11 +117,21 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
...
+    /**
+     * After HMR is triggered `SheetsManager` is resetted,
+     * so we have to be carefull here.
+     * HMR executes `createHoc`, `componentWillReceiveProps`, `componentWillUpdate` and `render`
+     * So here you can't unmanage sheet by theme, because manager is empty.
+     */
     componentWillUpdate(nextProps, nextState) {
-      if (isThemingEnabled && this.state.theme !== nextState.theme) {
+      const isThemeUpdate = isThemingEnabled && this.state.theme !== nextState.theme
+      const isHmrUpdate = process.env.NODE_ENV !== 'production' && (new Error()).stack.indexOf('hotReplacementRender') >= 0
```
kof: Wow, a hell of a trick

src/createHoc.js
```diff
@@ -202,11 +220,11 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
     }
 
     get jss() {
-      return this.context[ns.jss] || optionsJss || jss
+      return (this.context && this.context[ns.jss]) || optionsJss || jss
     }
 
     get manager() {
-      const managers = this.context[ns.managers]
+      const managers = this.context && this.context[ns.managers]
```
kof: When is it possible that this.context is not defined?

src/tests/hot-reloading.js
```diff
@@ -0,0 +1,43 @@
...
+  let ComponentB
+  let ComponentC
+
+  beforeEach(() => {
+    ComponentA = injectSheet({button: {color: 'red'}})()
+    ComponentB = injectSheet({button: {color: 'green'}})()
+    ComponentC = injectSheet({button: {color: 'blue'}})()
+  })
+
+  it.only('should hot reload component and attach new sheets', () => {
```
kof: don't forget this one

kof:
@felthy thanks for giving it a go, lets gather some feedback on this before we move forward

kof:
@felthy additional idea that came into my mind: what if we had a more universal API to trigger such updates? so that react-jss code base does not have to know of hmr existence and it can be triggered from outside? We could still have a module here or eventually even a separate package later, just stronger decoupling react-jss from hot reload would be nice, since this code is unrelated to what react-jss does and that is not nice to have here.

src/tests/hot-reloading.js
felthy: Not sure what you mean. Did you mean the second test from before, `should properly detach sheets on hot reloaded component`?

I omitted that one because I couldn’t tell the difference between the two tests, and also because "properly detach sheets" doesn’t actually happen. I spent a bit of time trying to find a way to get the old sheets to detach but I didn’t find one. I might have another look tomorrow but it would be great if anyone with more experience with this HMR stuff has any suggestions!

kof: you are using .only :) it will skip all the other tests and only run this one

felthy: Oh! I didn’t know what that meant, copied it over from before 😆

felthy: Yikes, 14 tests failing now, I’ll look into those tomorrow

kof: 😄

felthy:
Closing in favour of #123.

Issue #267, opened 2018-07-03
Static rules generate empty CSS rules in dynamic style sheet when a function rule is used


wouter-willems:
Hi,
whenever i am using dynamic properties, like:

```
wrapper: (props) => ({
    padding: 40,
    background: theme.background,
    textAlign: "left"
  }),
```

It causes an extra class to be added, which is empty and does nothing. You can see an example in this playground for the wrapper:

https://codesandbox.io/s/j3l06yyqpw

the result: `class="Comp-wrapper-0-1-10 Comp-wrapper-0-1-7"`

In that case, `Comp-wrapper-0-1-7` does nothing. You can see that even all children now have double classes.
This also does not work for the newest version, 9.8.7.

kof:
I see this: and I see that empty CSS rules are generated in the dynamic style instance: title and link classes should not be there, since they are fully static and are already rendered in the previous style.

<img width="340" alt="screen shot 2018-07-03 at 12 56 12 pm" src="https://user-images.githubusercontent.com/52824/42215992-8b450c18-7ec0-11e8-9fd1-6f9f740ea6fb.png">

wouter-willems:
What about the duplicate 'wrapper' class that is injected onto the dom element?

kof:
Oh yes thats a bug too, there is no even such CSS rule generated.

kof:
It is all related to function rules, I recommend to not use them for now and use function values only for dynamic values where really needed, since they don't support full syntax either, here is the issue https://github.com/cssinjs/jss/issues/682

wouter-willems:
Great, thank you kof! Looking forward to the fixes =)

HenriBeck:
For the rules in the dynamic sheet we could just remove any rule which is an emtpy object.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kof:
I think there is a bug in the dynamic rules extraction, there should be no rules which are completely static in the dynamic sheet.

HenriBeck:
Yeah, so it's an issue of getDynamicStyles function in jss. But we currently use that "bug" for the classes prop. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

HenriBeck:
The empty style rules are caused because we are depending here on `jss-compose` when we have styles with dynamic values. I think this should be removed.

kof:
[...]

PR #271, opened 2018-07-05
Fix having empty classes in the dynamic sheet

HenriBeck:
As discussed in #267 
This should remove empty classes and remove depending on jss-compose and jss-nested.

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -22,11 +22,13 @@
  * @return {Object|null}
  */
 export default (staticClasses, dynamicClasses) => {
+  const combinedClasses = {...staticClasses}
+
   for (const name in dynamicClasses) {
-    staticClasses[name] = staticClasses[name]
+    combinedClasses[name] = staticClasses[name]
       ? `${staticClasses[name]} ${dynamicClasses[name]}`
       : dynamicClasses[name]
   }
 
-  return staticClasses
+  return combinedClasses
 }
--- a/src/compose.js
+++ b/src/compose.js
@@ -21,32 +21,12 @@
  * @param {Object} styles dynamic styles object without static properties
  * @return {Object|null}
  */
-export default (classes, styles) => {
-  // Add `composes` property to rules which are already defined in `classes`.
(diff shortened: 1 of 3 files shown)


src/createHoc.js
```diff
@@ -146,9 +145,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
         })
       }
 
-      const sheet = dynamicSheet || staticSheet
       const defaultClasses = InnerComponent.defaultProps ? InnerComponent.defaultProps.classes : {}
-      const classes = {...defaultClasses, ...sheet.classes, ...userClasses}
+      const classes = {
+        ...defaultClasses,
+        ...compose(staticSheet.classes, dynamicSheet ? dynamicSheet.classes : {}),
```
kof: if no dynamicSheet why calling compose at all?

HenriBeck: Yeah, added that improvement locally, just forgot to push

src/compose.js
```diff
@@ -21,32 +21,12 @@
...
-    // Add rules which are defined in `classes` but aren't in styles.
-    for (const name in classes) {
-      const className = styles[name]
-      if (!className) {
-        styles[name] = {composes: classes[name]}
-      }
-    }
+export default (staticClasses, dynamicClasses) => {
+  for (const name in dynamicClasses) {
+    staticClasses[name] = staticClasses[name]
```
kof: you are mutating the classes object that you got from the static style sheet, we shouldn't mutate that

src/createHoc.js
```diff
@@ -146,9 +145,15 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
         })
       }
 
-      const sheet = dynamicSheet || staticSheet
       const defaultClasses = InnerComponent.defaultProps ? InnerComponent.defaultProps.classes : {}
```
kof: btw instead of empty object it can be undefined, spreading still works (learned it after I wrote it I suppose)

PR #279, opened 2018-07-16
Added flow typings

gomesalexandre:
Hi ! Here you'll find the flow typings for the project.
- I've also added a `.flowconfig` file, and a `react/sort-comp` eslint rule so types can be defined at the top.
- `build:lib` script has been updated with a `flow-copy-source` step, so that `.js.flow` files are generated on the lib folder, next to their `.js` counterpart.

I've tried to dig into the code and understand the shape of the data structures as much as possible, but if you feel some types are too generic, too strict, or are missing some fields, you tell me 😉


src/types.js
```diff
@@ -0,0 +1,16 @@
...
+  theming?: {
+    themeListener: any => any
+  },
+  inject?: [],
+  jss?: any,
+  index?: ?number
+};
+export type Theme = {};
+export type Styles = {[string]: {}};
+export type ThemerFn = (theme: Theme) => {};
```
HenriBeck: The ThemerFn should return Styles

src/createHoc.js
```diff
@@ -71,10 +94,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
   const noTheme = {}
   const managerId = managersCounter++
   const manager = new SheetsManager()
-  const defaultProps = {...InnerComponent.defaultProps}
+  const defaultProps: DefaultProps = {...InnerComponent.defaultProps}
   delete defaultProps.classes
 
-  class Jss extends Component {
+  class Jss extends Component<Props, State> {
+    state: State;
```
HenriBeck: The state doesn't need to be typed here again.

src/createHoc.js
```diff
@@ -86,7 +111,7 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
     }
     static defaultProps = defaultProps
 
-    constructor(props, context) {
+    constructor(props: Props, context: ComponentType<Props>) {
```
HenriBeck: Wrong definition. The context is not a Component.

src/createHoc.js
```diff
@@ -103,13 +128,13 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       }
     }
 
-    componentWillReceiveProps(nextProps, nextContext) {
+    componentWillReceiveProps(nextProps: Props, nextContext: ComponentType<Props>) {
```
HenriBeck: Usually, flow should infer the types for the lifecycle methods.

src/createHoc.js
```diff
@@ -183,7 +208,7 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       return {theme, dynamicSheet, classes}
     }
 
-    manage({theme, dynamicSheet}) {
+    manage({theme, dynamicSheet}: {theme: Theme, dynamicSheet: DynamicSheet}) {
```
HenriBeck: What `manage` actually get's is the state, you should be able to just say `State` there

src/createHoc.js
```diff
@@ -61,7 +84,7 @@ let managersCounter = 0
  * @param {Object} [options]
  * @return {Component}
  */
-export default (stylesOrCreator, InnerComponent, options = {}) => {
+export default (stylesOrCreator: StylesOrThemer, InnerComponent: any, options: Options = {}) => {
```
HenriBeck: We should be able to type the InnerComponent better. Take a look at https://flow.org/en/docs/react/hoc/

src/createHoc.js
```diff
@@ -32,7 +34,28 @@ const dynamicStylesNs = Math.random()
  *
  */
 
-const getStyles = (stylesOrCreator, theme) => {
+type DynamicSheet = {
```
kof: we should use StyleSheet from jss

src/types.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+export type Options = {
```
kof: this should also extend the jss style sheet options with additional options

src/createHoc.js
gomesalexandre: My bad, I see that Facebook uses any for context, I'll use the same here 
`(props: Props, context: any): React$Node,` taken from `react.js` flow libdef

src/createHoc.js
TrySound: Nope. Flow requires types for public methods of exported class.

gomesalexandre: Indeed @TrySound , if not annotating type of a lifecycle method parameter i.e `nextProps` on `componentWillReceiveProps` flow will throw this:
`Missing type annotation for nextProps.`

src/types.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import {JssSheet} from './jss'
+
+export type Options = {
```
HenriBeck: The options also get a `theming` object, a `inject` property which is an array o 'classes', 'theme', 'sheet' and a jss option which is a Jss Instance.

src/injectSheet.js
```diff
@@ -24,12 +31,13 @@ const NoRenderer = ({children}) => (children || null)
  *
  * @api public
  */
-export default function injectSheet(stylesOrSheet, options = {}) {
+export default function injectSheet(stylesOrSheet: StylesOrThemer, options: Options = {}) {
   if (options.index === undefined) {
     options.index = indexCounter++
   }
-  return (InnerComponent = NoRenderer) => {
+  function injector<P: any>(InnerComponent: React$ComponentType<P> = NoRenderer) {
```
HenriBeck: You should use `React.ComponentType` instead of `React$ComponentType`. Also, the Props must always be an object, so you should be able to do `P: {}`

src/createHoc.js
```diff
@@ -71,10 +89,11 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
   const noTheme = {}
   const managerId = managersCounter++
   const manager = new SheetsManager()
-  const defaultProps = {...InnerComponent.defaultProps}
+  const defaultProps: DefaultProps = {...InnerComponent.defaultProps}
```
HenriBeck: We could type the default props just as the props of the component, I think. I guess we need that typing for the next line to work.

src/types.js
gomesalexandre: Are any of the array elements optional ?
Also, should the `jss` field override the one from `StylesheetOptions` (in jss)?

src/createHoc.js
gomesalexandre: Given `classes` is the only required field in Props, sure, that would work. Do you think it would be a good idea to remove DefaultProps ? I chose to have a separate one for readability.

src/createHoc.js
TrySound: Yes. You should annotate.

src/types.js
HenriBeck: There is no `jss` option inside the StyleSheet Options.

just type the array like that: `Array<'classes' | 'theme' | 'sheet'>`

That should only allow those values inside the array.

src/injectSheet.js
HenriBeck: Can you try making the `children` optional and not the value?

src/createHoc.js
```diff
@@ -103,13 +122,13 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       }
     }
 
-    componentWillReceiveProps(nextProps, nextContext) {
+    componentWillReceiveProps(nextProps: Props, nextContext: ComponentType<Props>) {
```
HenriBeck: The context will definitely not be a `Component`. For now, maybe just type it as an Object and we can add safer typings later.

gomesalexandre: Will do ! Typed both `context` and `nextContext` as objects for now

src/injectSheet.js
gomesalexandre: In that case I get this : 
```javascript
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/injectSheet.js:25:20

Property children is missing in object type [1] but exists in NoRendererProps [2] in the first argument.

     src/injectSheet.js
 [2]  25│ const NoRenderer = ({children}: NoRendererProps) => (children || null)
        :
      35│   if (options.index === undefined) {
      36│     options.index = indexCounter++
      37│   }
      38│   function injector<P: {}>(InnerComponent: React.ComponentType<P> = NoRenderer) {
      39│     const Jss = createHoc(stylesOrSheet, InnerComponent, options)
      40│     return hoistNonReactStatics(Jss, InnerComponent, {inner: true})
      41│   }

     /private/tmp/flow/flowlib_2f2515e2/react.js
 [1] 138│   (props: Props, context: any): React$Node,
```

I managed to fix this one by diff-ing NoRendererProps and the generic P

src/createHoc.js
gomesalexandre: Please have a look at he most recent commit, I've added better HOC typing

src/types.js
gomesalexandre: Fixed

src/createHoc.js
HenriBeck: Which commit? Your last commit was 2 days ago, did you forgot to push?

gomesalexandre: My bad, it wasn't passing the pre-hooks tests. Have a look now (please note that using Props as a generic with createHOC function is still a WIP currently)

gomesalexandre:
@HenriBeck  jss defines the 'update' method as taking a `name` parameter of type `string`, however in react-jss usage, this is what we have : 
```
Cannot call dynamicSheet.update with this.props bound to name because Props [1] is incompatible with string [2].

     src/createHoc.js
 [1]  97│   class Jss extends Component<Props, State> {
        :
     217│
     218│       if (dynamicSheet) {
     219│         dynamicSheet
     220│           .update(this.props)
     221│           .attach()
     222│         if (registry) registry.add(dynamicSheet)
     223│       }
```

In our duplicated version of StyleSheet, should we make name an union of types `string | Props` instead ?

kof:
@gomesalexandre no first argument can be a string, or an object, if its a string its a rule name and second argument is the data object. Its called an arguments overload pattern.

kof:
see https://github.com/cssinjs/jss/blob/master/packages/jss/src/RuleList.js#L8

gomesalexandre:
@kof This fixed it indeed, but now I got errors coming from `jss` in `node_modules` : 
```javascript
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/jss/lib/RuleList.js.flow:145:24

Cannot call plugins.onUpdate with name bound to data because undefined [1] is incompatible with object type [2].

     node_modules/jss/lib/RuleList.js.flow
 [1] 137│   update(name?: string | Object, data?: Object): void {
        :
     142│     }
     143│
     144│     for (let index = 0; index < this.index.length; index++) {
     145│       plugins.onUpdate(name, this.index[index], sheet)
     146│     }
     147│   }
     148│

     node_modules/jss/lib/PluginsRegistry.js.flow
 [2]  72│   onUpdate(data: Object, rule: Rule, sheet: StyleSheet): void {
```

This is gone after refinement : 

```javascript
  update(name?: string | Object, data?: Object): void {
    const {jss: {plugins}, sheet} = this.options
    if (name && typeof name === 'string') {
      if (data) {
        plugins.onUpdate(data, this.get(name), sheet)
      }
      return
    }
    
    for (let index = 0; index < this.index.length; index++) {
      if (name) {
        plugins.onUpdate(name, this.index[index], sheet)
      }
    }
  }
```

HenriBeck:
@gomesalexandre you shouldn't change code inside node_modules.

You can set the whole node_modules folder to only use declarations and skip errors:
https://flow.org/en/docs/config/declarations/

kof:
yes, lets do this

HenriBeck:
Okay, will see if I can move it tomorrow. I'm sorry @gomesalexandre but you will have to recreate your PR.

gomesalexandre:
Sure @HenriBeck I will. I'll push some changes that I made in the meantime

HenriBeck:
@gomesalexandre, please recreate the PR in the `jss` repo and create your branch based on the `add-react-jss` that I just created. You can import the types you need from `jss`

gomesalexandre:
Good @HenriBeck I'll create it tomorrow, with whatever changes I need to make following the switch to the monorepo

PR #281, opened 2018-07-18
Fix `Cannot read property 'detach' of undefined`

felthy:
The only reason `react-jss-hmr` patches `componentDidUpdate()` is to stop this code from throwing when a dynamic sheet gets added via HMR. 

If we add this check, I can merge cssinjs/react-jss-hmr#3 and the code over there will be simpler.

On the other hand, this check is completely unnecessary except when `react-jss-hmr` is being used.


HenriBeck:
I would say it actually makes sense as it makes clear that we only remove the sheet when it actually exists.

PR #314, opened 2018-11-28
Fix hoisting of new react features

eps1lon:
`hoist-non-react-statics@2.x` does not support `forwardRef` components, `contextType` static nor `getDerivedStateFromError` static. The major version bump was due to the added requirement of react >= 0.14 which was dropped in `hoist-non-react-statics@3.2.0`. Therefore this change should not be breaking.

[`hoist-non-react-statics` changelog](https://github.com/mridgway/hoist-non-react-statics/blob/master/CHANGELOG.md)


kof:
this package has migrated to jss repository monorepo, please resubmit there

